PR #7, opened 2021-12-14
Add logger, docs, checks and fix flake8

mscheltienne:
First PR to familiarize myself with your code.

Looks really good and solid; I added in `utils` 3 files:
- `_checks.py` that contains functions to check user input in the public API
- `_docs.py` that contains the `fill_doc` decorator and the `docdict` for this library. I added a couple of lines to load entries from MNE's `docdict` to avoid duplication.
- `_logs.py` that contains the logger and associated functions.

I added a flake8 style workflow and I've started fixing the PEP8/Flake8 style issues.

I'll replace the logger across the code-base with the logger from `_logs.py`, I'll add probably also a `verbose` decorator similar to the one from MNE since you used it a lot.

Also, I've changed a little `setup.py` because it was not loading requirements from the `.txt` file. Why is the version of MNE restricted to 0.22 max? I think this should be compatible with new versions of MNE as well.


mscheltienne:
I broke more stuff than I was expecting to 😅 
I'll fix the remaining couple of typos later.

vferat:
With latest version of mne, tests break because the pooch library is need to download datasets.
`RuntimeError: For dataset downloading to work, the pooch module is needed, but it could not be imported`

You can try to add pooch to both ` requirements_testing.txt ` and `requirements_doc.txt` and change `mne <= 0.22.0` to `mne` in ` requirements_testing.txt `, `requirements_doc.txt` , `requirements.txt`.

Otherwise I'll do it after merging this PR.

vferat:
@all-contributors please add @mscheltienne  for code

vferat:
@all-contributors please add @mscheltienne for code

PR #13, opened 2022-02-09
Fix segment

vferat:
Fix predict algorithm


mscheltienne:
I've copied the reference in #14, so if we go ahead with #14; this PR is not needed.

PR #14, opened 2022-02-09
Refactor clustering

mscheltienne:
PR to propose a refactor structure for clustering algorithm(s):

- Define a base abstract private class `_BaseCluster`: what is a clustering class suppose to look like, what are the common variable/methods, what are the abstract method that are expected to be present, ...
- Opens the door for different clustering algorithms to be added in the (far) future.
- Rename `clustering` to `cluster` and `ModKMeans` to `KMeans` to match scikit-learn: https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html   What did `Mod` stands for anyway?
- Add more error checking on user input, add key variables as read-only properties
- Split properties between pre-fit (defines the algorithm) and post-fit, have the post-fit variables reset if `.fitted` is set to False.
- Improve some docstrings
- Improve some parts of the code


pycrostates/cluster/kmeans.py
```diff
@@ -0,0 +1,289 @@
...
+    def random_state(self):
+        """
+        Random state.
+
+        :type: `None` | `int` | `~numpy.random.RandomState`
+        """
+        return self._random_state
+
+    @property
+    def GEV(self):
```
vferat, line 229: Maybe we can follow the sklearn convention for this one  -> Attributes that have been estimated from the data must always have a name ending with trailing underscore
def GEV_(self):

and change self._GEV to self.GEV_

pycrostates/cluster/kmeans.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+        GEV fit variable.
+        """
+        if self._GEV is None:
+            assert not self._fitted  # sanity-check
+            logger.warning('Clustering algorithm has not been fitted.')
+        return self._GEV
+
+    @property
+    def labels(self):
```
vferat, line 239: sklearn convention:   _labels -> labels_

pycrostates/cluster/kmeans.py
mscheltienne: Ok, so I'll keep the attribute as `self._gev` (private) but rename the property to `GEV_`

pycrostates/cluster/kmeans.py
mscheltienne: What is the difference? ^^' What is the sklearn convention for this one?

pycrostates/cluster/kmeans.py
mscheltienne: To be sure, there are 3 variables that have been estimated from data and should end with a trailing underscore: `GEV`, `labels` and `cluster_centers`. Correct?

mscheltienne:
[52772f0](https://github.com/vferat/pycrostates/pull/14/commits/52772f0b0ff284713e9a829560def3480adafae9) passes current tests.

PR #19, opened 2022-03-14
Update _BaseCluster.reorder() to also reorder _BaseCluster._labels_

vferat:


pycrostates/cluster/_base.py
```diff
@@ -270,6 +270,11 @@ def reorder_clusters(self, mapping: dict = None,
         # re-order
         self._cluster_centers_ = self._cluster_centers_[order]
         self._clusters_names = [self._clusters_names[k] for k in order]
+        new_labels = np.zeros(self._labels_.shape)
```
mscheltienne, line 273: Should be in kmeans.py since it's an attribute from kmeans class only.
```
def reorder_clusters(...):
    super().reorder_clusters()
    # additional steps
    ...
```

pycrostates/cluster/_base.py
```diff
@@ -747,6 +753,16 @@ def fitted(self, fitted):
             self._fitted_data = None
             self._fitted = False
```
mscheltienne, line 754: Add the reset for `self._labels_` in the base class (and remove from `fitted` setter in KMeans.

pycrostates/cluster/_base.py
```diff
@@ -270,6 +271,11 @@ def reorder_clusters(self, mapping: dict = None,
         # re-order
         self._cluster_centers_ = self._cluster_centers_[order]
         self._clusters_names = [self._clusters_names[k] for k in order]
+        if self._labels_ is not None:
+            new_labels = np.zeros(self._labels_.shape)
+            for k in range(0, self.n_clusters):
+                new_labels[self._labels_ == k+1] = order[k] + 1
+            self._labels_ = new_labels
```
mscheltienne, line 278: `order` is 0-indexed (to be added to the docstring). So if I try this snippet with 4 clusters:

```
import numpy as np


order = [3, 1, 2, 0]
labels = np.array([0, 1, 1, 2, 3, 2, 1, 0, 0])

new_labels = np.zeros(labels.shape)
for k in range(0, 4):
    new_labels[labels == k+1] = order[k] + 1
```

I should get `[3, 1, 1, 2, 0, 2, 1, 3, 3]` since 0 and 3 have swap. With this loop I get: `array([0., 4., 4., 2., 3., 2., 4., 0., 0.])`. Am I missing something?

So maybe something like that:

```
new_labels = labels.copy()
for old, new in enumerate(order):
    new_labels[labels == old] = new
```

pycrostates/cluster/tests/test_kmeans.py
```diff
@@ -293,6 +293,12 @@ def test_reorder(caplog):
     assert ModK.clusters_names[0] == ModK_.clusters_names[1]
     assert ModK.clusters_names[0] == ModK_.clusters_names[1]
 
+    # test .labels_ reordering
+    x = ModK_.labels_[:20]
+    # x = [3, 1, 1, 2, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0]
+    y = [3, 2, 2, 1, 2, 2, 2, 2, 2, 2, 2, 2, 0, 0, 0, 0, 0, 0, 0, 0]
```
mscheltienne, line 299: So following my comment above that order is 0-indexed, `ModK_.reorder_clusters(order=np.array([1, 0, 2, 3]))` should:
- Place cluster 1 in position 0
- Place cluster 0 in position 1
- Keep cluster 2 in position 2
- Keep cluster 3 in position 3

Thus expected labels should be: `[3, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1]`.

pycrostates/cluster/_base.py
vferat: As for now` .labels_ ` uses 0 for unlabeled data, 1 for the first cluster_centers ( i.e. cluster_centers [0]) and so on...
But you spot a mistake as their should be no 0 in `.labels_` 
[this ligne](https://github.com/vferat/pycrostates/blob/615120870e1642318e0c2a100599fff08e4558e2/pycrostates/cluster/kmeans.py#L108)  `self._labels_ = best_segmentation` should be changed to  `self._labels_ = best_segmentation + 1`

PR #21, opened 2022-03-16
Add ChInfo class to hold channel-related information only

mscheltienne:
- [x] Add ChInfo class (created either from ch_names/ch_types or from an Info)
- [ ] Add support for picking channels [not at info level]
- [ ] Add support for renaming/reordering channels [not at info level]
- [x] Add support for getting/setting montage
- [x] Add tests

EDIT: Not implementing channel picking, renaming, reordering as an info instance does not have any information about the underlying data array. Thus, those methods should be only implemented in the class that has a `.info` attribute and a `.data` attribute with the underlying data reordered/picked.


pycrostates/cluster/_base.py
```diff
@@ -327,8 +327,8 @@ def plot(self, block=False):
 
         Returns
         -------
-        fig : :class:`matplotlib.figure.Figure`
-            Figure
+        fig : Figure
```
mscheltienne, line 330: Figure, as for all arguments in `Parameters` / `Returns` should be resolved by `numpydoc` directly.

pycrostates/cluster/_base.py
```diff
@@ -327,8 +327,8 @@ def plot(self, block=False):
...
+            Matplotlib figure containing the topographic plots.
```
mscheltienne, line 331: Figure, as for all arguments in `Parameters` / `Returns` should be resolved by `numpydoc` directly.

pycrostates/utils/_docs.py
```diff
@@ -11,21 +11,29 @@
...
 
 # ---- Clusters ----
 docdict['fit_inst'] = """
-inst : `~mne.io.Raw` | `~mne.Epochs`
+inst : Raw | Epochs
     Instance containing data to transform to cluster-distance space
     (absolute spatial correlation)."""
 docdict['predict_inst'] = """
-inst : `~mne.io.Raw` | `~mne.Epochs`
+inst : Raw | Epochs
```
mscheltienne, line 36: Same as for Figure, Raw, Epochs, those names should be resolved by numpydoc directly.

mscheltienne:
That should be enough for the class itself. I think integrating this class with our existing code-base should be done in a separate PR. WDYT?

vferat (approved):

PR #27, opened 2022-04-05
Improvements to viz

mscheltienne:
Code clean-up to viz + additional functionalities and error checking (for cluster_centers)

### DIFF
--- a/pycrostates/viz/cluster_centers.py
+++ b/pycrostates/viz/cluster_centers.py
@@ -17,27 +17,12 @@ def plot_cluster_centers(
         **kwargs,
         ):
     """
-    Create topographic plot for cluster_centers.
-
-    Parameters
-    ----------
-    cluster_centers : TYPE
-        DESCRIPTION.
-    info : TYPE
-        DESCRIPTION.
-    cluster_names : TYPE, optional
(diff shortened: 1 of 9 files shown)


pycrostates/viz/cluster_centers.py
```diff
@@ -0,0 +1,92 @@
...
+                "number of clusters and Axes.")
+        figs = [ax.get_figure() for ax in axes]
+        if len(set(figs)) == 1:
+            f = figs[0]
+        else:
+            f = figs
+        del figs
+
+    # remove show from kwargs, issue warning if it is present
+    if 'show' in kwargs:
```
vferat, line 72: We could use something like to keep 'show' in kwargs
 ```
kwargs = {key: value for key, value in kwargs.items()
                 if key not in ('show', )}


for k, (center, name) in enumerate(zip(cluster_centers, cluster_names)):
    ...
    plot_topomap(center, info, axes=ax, show=False, **kwargs)
    ax.set_title(name)

if 'show' in kwargs:
    if kwargs['show']:
        plt.show(block=block)
return f
```

mscheltienne: good idea, done

PR #31, opened 2022-04-13
Add pypi publish workflow

vferat:
Add github workflow to test deployment on Pypi and automatically deploy on new tag

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yml
+++ b/.github/workflows/publish-to-test-pypi.yml
@@ -1,0 +1,37 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+# https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+on: push
+
+jobs:
+  build-n-publish:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-to-test-pypi.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+# https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+on: push
```
mscheltienne, line 3: Should be:

```
on:
  release:
    types: [published]
```

mscheltienne:
Have a look at this one: https://github.com/bsl-tools/bsl/blob/main/.github/workflows/publish.yml
I don't remember if I tested it or not, but it's simpler

mscheltienne:
Although this workflow looks awesome and better 👍

mscheltienne:
FYI, the steps to make an MNE release: https://github.com/mne-tools/mne-python/wiki/How-to-make-a-release
I think we should be OK with a couple of manual steps 😅 Can we close this PR?

PR #35, opened 2022-05-02
Use black and isort for formating

mscheltienne:
Proposal to use [black](https://black.readthedocs.io/en/stable/) and [isort](https://github.com/PyCQA/isort) to format the code.
I just added the configuration for now.


.github/workflows/code-style.yml
```diff
@@ -37,3 +37,18 @@ jobs:
...
+        uses: psf/black@stable
+        with:
+          options: "--check --verbose"
+      - name: Run isort
+        uses: jamescurtin/isort-action@master
+        with:
+            requirementsFiles: "requirements.txt requirements_testing.txt"
+      - name: Install pydocstyle
+        run: | 
+          python -m pip install -r requirements_testing.txt
```
mscheltienne, line 50: Line 30:33 Install flake8
If you do have flake8 in `requirements_testing.txt`. which you should, then instead of doing a second installation step just rename to `Install dependencies` and uses this earlier instead of `Install flake8`. You can also add the upgrade of pip, setuptools and wheel and the `--progress-bar off` to reduce verbovsity.

.github/workflows/code-style.yml
```diff
@@ -27,10 +27,11 @@ jobs:
         with:
           python-version: '3.9'
           architecture: 'x64'
-      - name: Install flake8
+      - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip setuptools wheel
-          python -m pip install flake8
+          python -m pip install --upgrade pip setuptools wheel --progress-bar off
+          python -m pip install requirements.txt
```
mscheltienne:
```suggestion
```

mscheltienne: we don't need those req. for style formatting, do we?

.github/workflows/code-style.yml
```diff
@@ -27,13 +27,25 @@ jobs:
         with:
           python-version: '3.9'
           architecture: 'x64'
-      - name: Install flake8
+      - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip setuptools wheel
-          python -m pip install flake8
+          python -m pip install --upgrade pip setuptools wheel --progress-bar off
+          python -m pip install -r requirements_testing.txt
```
mscheltienne, line 33:
```suggestion
          python -m pip install -r requirements_testing.txt --progress-bar off
```

mscheltienne:
@vferat Merge if happy.

PR #37, opened 2022-05-16
Dev metrics

vferat:


vferat:
Code for quantitative tests:
https://gist.github.com/vferat/ad463d439d9d351f7b3a5c7b83f10dc7

pycrostates/metrics/metrics.py
```diff
@@ -61,13 +65,72 @@ def silhouette(modK):  # lower the better
...
+def calinski_harabasz(cluster):  # higher the better
+    """Compute the Calinski and Harabasz score.
+
+    This function is a proxy function for
+    :func:`sklearn.metrics.calinski_harabasz_score` that applies directly to a
+    fitted :class:`pycrostate.clustering.BaseClustering`.
+
+    Parameters
+    ----------
+    cluster : :class:`pycrostate.clustering.BaseClustering`
```
mscheltienne, line 78: Those x-links will not work as it's `pycrostates.clustering._base._BaseClustering` and as it should not be documented. 
Maybe it becomes public and documented, with a warning pointing towards the actual instantiable classes (e.g. ModK), similarly to https://mne.tools/dev/generated/mne.Info.html
Alternatively, we could define a `:ref:` to a page that details what a clustering solution is, what it contains and what are the different algorithm available.
I think I prefer option 2.

vferat: I also prefer option 2

tutorials/metrics/plot_metrics.py
```diff
@@ -0,0 +1,96 @@
...
+raw.pick("eeg")
+raw.set_eeg_reference("average")
+
+#%%
+# In this example, we will used the
+# [Silhouette Coefficient](https://doi.org/10.1016/0377-0427(87)90125-7) to
+# evaluate the
+# Pycrostates implementations relies on sklearn implementation
+# while ensuring that the metric used for distance computations is
+# the opposite value of the absolute spatial correlation.
```
mscheltienne, line 35: Something is missing in this sentence.

mscheltienne (commented):
I re-structured to split the metrics in different files and benefit from the module structure. If you prefer to keep it all in one file, you can revert the last commit.
Before this last commit, I fixed some typos and some remaining x-ref to `BaseClustering`.
If you revert the last commit; please double-check the docstring of each metrics, some of them had `"The davies bounding score"` as a return instead of the correct name, and I fixed it as part of the re-structure commit.

tutorials/metrics/plot_metrics.py
```diff
@@ -0,0 +1,106 @@
...
+# Pycrostates implements several metrics to evaluate the quality of
+# clustering solutions without knowing the ground truth:
+# - The silhouette score (higher the better)
+# - The Calinski Harabasz score (higher the better)
+# - The Dunn score (higher the better)
+# - The Davies Bouldin score (lower the better)
+#
+# We can apply these metrics on clustering solution
+# computed with different values of n_clusters and
+# analyse which give the best clsutering solution.
```
mscheltienne, line 39:
```suggestion
# analyse which give the best clustering solution. 
```

mscheltienne (approved):

PR #39, opened 2022-05-27
[MNT] Move package configuration and dependencies to pyproject.toml (PEP 518, 621)

mscheltienne:
I started looking into new ways to package python libraries. The direction seems to be to put everything in `pyproject.toml`, thus getting rid of `requirements.txt` files and of `setup.py` or `setup.cfg`.
`setup.py` is still needed to install the package in editable mode until setuptools complies with PEP 660, but it can be a simple wrapper as i this PR while all the metadata are in `pyproject.toml`.

I updated the workflows accordingly to work with this different packaging.
I added a workflow to test the build, using [build](https://pypa-build.readthedocs.io/en/stable/) on all the supported platforms.
I added a codespell action to check spelling, and fixed all the typos it found.
I updated the regex pattern for pydocstyle, and added the properties to the ignored decorators.

Let's see how badly it brokes the CIs or anything :sweat_smile:


pycrostates/datasets/lemon/lemon.py
```diff
@@ -2,15 +2,14 @@
...
-from mne.io import BaseRaw
+from mne import create_info
+from mne.io import BaseRaw, RawArray
 
 from ...utils._checks import _check_type, _check_value
 from ...utils._config import get_config
-from ...utils._imports import import_optional_dependency
 
 
 def load_data(subject_id: str, condition: str):
```
mscheltienne, line 15: Since it returns a path (as a string), I think it would be better to match MNE API by naming this function `data_path` and returning a `pathlib.Path` instance.

pycrostates/datasets/lemon/lemon.py
```diff
@@ -100,86 +92,35 @@ def standardize(raw: BaseRaw):
...
+        ch_names = raw.info["ch_names"] + missing_channels
+        ch_types = raw.get_channel_types() + ["eeg"] * len(missing_channels)
+        info = create_info(
+            ch_names=ch_names, ch_types=ch_types, sfreq=raw.info["sfreq"]
+        )
+        raw = RawArray(data=data, info=info)
+        raw.info["bads"].extend(missing_channels)
+
     raw.reorder_channels(standard_channels)
     raw.set_montage("standard_1005")
```
mscheltienne, line 125: Before setting the montage and re-referencing to a CAR, we could add back the reference channel (if we know which one was used) with `raw.add_reference_channels`.

pycrostates/datasets/lemon/lemon.py
```diff
@@ -36,26 +36,19 @@ def load_data(subject_id: str, condition: str):
     Returns
     -------
     path : str
```
mscheltienne, line 38:
```suggestion
    path : Path
```

pycrostates/datasets/lemon/lemon.py
```diff
@@ -36,26 +36,19 @@ def load_data(subject_id: str, condition: str):
...
-        path to local copy of the requested recording .
+        path to local copy of the requested recording.
```
mscheltienne, line 39:
```suggestion
        Path to a local copy of the requested recording.
```

pycrostates/datasets/lemon/lemon.py
```diff
@@ -1,19 +1,19 @@
...
+from mne import create_info
+from mne.io import BaseRaw, RawArray
 
 from ...utils._checks import _check_type, _check_value
 from ...utils._config import get_config
-from ...utils._imports import import_optional_dependency
 
 
-def load_data(subject_id: str, condition: str):
+def data_path(subject_id: str, condition: str):
```
mscheltienne, line 16:
```suggestion
def data_path(subject_id: str, condition: str) -> Path:
```

vferat (approved):

PR #41, opened 2022-06-07
Try to fix test CI

mscheltienne:


mscheltienne:
I'm still struggling with the random Tcl error. It occurs only in windows build, on all 3 python versions. 
For some reason, there are 2 different errors raised:
- one where the tcl init is not found
- one where `invalid command name "tcl_findLibrary"` is raised
For some reason, the tests don't fail at the same point and it usually fails on a `plt.subplots` command. And finally, for some reason... it can fail on a `plt.subplots` command even if previous `plt.subplots` were executed fine. e.g.

```
    def test_plot_epoch_segmentation():
        """Test segmentation plots for epochs."""
        n_clusters = 4
        labels = np.random.choice(
            [-1, 0, 1, 2, 3], (len(epochs), epochs.times.size)
        )
    
        plot_epoch_segmentation(labels, epochs, n_clusters)
        plt.close("all")
    
        # provide ax
        f, ax = plt.subplots(1, 1)
        plot_epoch_segmentation(labels, epochs, n_clusters, axes=ax)
        plt.close("all")
    
        # provide cbar_ax
>>      f, cbar_ax = plt.subplots(1, 1)
```

Anyway, I fixed other stuff in this PR in the meantime:
- I changed the one test that was using the MNE sample dataset to use the MNE testing dataset, to avoid downloading 2 datasets (-2.7 Gb of download on every job).
- I removed the call to `Path` on the MNE `.data_path()` function, as it already returns a `Path`
- I regrouped the 3 azure-pipeline files into one, as the MNE stable vs main version differed by only 2 lines, and as the build on the different OS differed only by the VM-image.
- Now, all the variables to control the pipeline are at the top of the azure-pipeline file.
- The codecov upload was also picking up the MNE dev pipeline, so I bumped this one to 3.10
- I added support for `[skip azp]` (or `[azp skip]`, or `[skip ci]` or `[ci skip]`) which is actually not supported by Azure for PR commits. 
- I added caching of the MNE testing dataset, with a key based on a str + the version of the dataset, to update the cache each time a new version is released.
- I fixed the doc-build, the commit 9d325d0d9e51d36c612b16cff810d7583fae04b3 was not enough.

I think we can already merge this, and I'll keep asking around for the Tcl issue..

vferat (approved):

PR #53, opened 2022-07-21
[DOC] Edit the versionadded and versionchanged flags from MNE docdict entries

mscheltienne:


mscheltienne:
I don't think the keys we used had a `versionadded` or `versionchanged` flag, but I ran into this today with `mne-icalabel`, so let's fix it here as well ;)

LGTM, merge if happy.

Issue #51, opened 2022-07-14
Review JOSS


tuliofalmeida:
Hi. This is an issue created as part of [pycrostates JOSS Review](https://github.com/openjournals/joss-reviews/issues/4564)

I've been doing some tests and following the examples you provided in the documentation. First, congratulations for the work, I liked the way you developed the package and I believe it will be very useful.

All the things I tried worked fine, so far so good. I have few things to suggest that are quite simple. I mostly used google colab to test the package and I missed your examples of adding the installation of the necessary packages.

I had to add the following lines in the examples:

```
!pip install -U mne 
!pip install pycrostates 
!pip install pymatreader 

import pymatreader 
```

The only example where I had to change some things to run was "plot_extract_gfp_peaks". In this I managed to solve the errors by performing the installation through your repository. Also the "dark mode" of the documentation some topics were hard to see (I don't know if it was a matter of my browser settings).

mscheltienne:
I added this note for the examples.

```
# .. note::
#
#     The lemon datasets is composed of EEGLAB files. To use the MNE reader
#     :func:`mne.io.read_raw_eeglab`, the ``pymatreader`` optional dependency
#     is required. Use the following installation method appropriate for your
#     environment:
#     - ``pip install pymatreader```
#     - ``conda install -c conda-forge pymatreader```
#     Note that an environment created via the MNE installers includes
#     ``pymatreader`` by default.
```

Also, if MNE is installed with the [standalone installers](https://mne.tools/stable/install/index.html), it will include all its optional dependencies and `pycrostates`, as of the next version 1.1.0 which should be released very shortly (https://github.com/mne-tools/mne-installers/pull/142).

tuliofalmeida:
Sorry for the delay! I agree with using the note, I think it will work fine and I didn't know about the pydata-sphinx-theme issue, sorry.

About the "plot_extract_gfp_peaks" after installing pycrostates and pymatreader I found this error:

`
from pycrostates.preprocessing import extract_gfp_peaks
gfp_data = extract_gfp_peaks(raw, min_peak_distance=3)
gfp_data

[extract_gfp_peaks.extract_gfp_peaks] INFO: 14713 GFP peaks extracted out of 119278 samples (12.34% of the original data).

---------------------------------------------------------------------------

TemplateNotFound                          Traceback (most recent call last)

[/usr/local/lib/python3.7/dist-packages/IPython/core/formatters.py](https://localhost:8080/#) in __call__(self, obj)
[...]

PR #54, opened 2022-07-26
Package jinja files in the tarball and the wheel

mscheltienne:
As noticed by @tuliofalmeida in #51


mscheltienne:
@vferat Can you make a new 0.1.2 release? It would be good to release this fix on `pip` and `conda` before `mne` 1.1.0 and its installers are released.

mscheltienne:
Done as MNE is about to bump the version today or tomorrow, it might even be too late to bump the version in the installers.

PR #66, opened 2022-08-20
Fix default picks

vferat:
Change the default  `BaseCluster.fit` parameter `picks `from "None" (all channels) to "eeg" as the current implementation is tested for eeg channels only.


mscheltienne:
OK to set the default to `eeg` as it's the current use case, but after a quick test on main, there is something else going on.

```
from mne.datasets import testing
from mne.io import read_raw_fif

from pycrostates.cluster import ModKMeans


directory = testing.data_path() / "MEG" / "sample"
fname = directory / "sample_audvis_trunc_raw.fif"
raw = read_raw_fif(fname, preload=False)
raw.pick_types(meg=True, eeg=True, eog=True)
raw.crop(0, 10)
raw.load_data()

ModK = ModKMeans(n_clusters=4, n_init=10, max_iter=100, tol=1e-4)
ModK.fit(raw, n_jobs=1)
```

If you run `ModK.plot()`, you get:

```
ValueError: Multiple channel types in Info structure. Pick Info with e.g. mne.pick_info and mne.io.pick.channel_indices_by_type.
```

------------

We only tested for the case where we provide a a raw with multiple channel types and provide a `picks` argument to select only one channel type. e.g. `ModK.fit(raw, picks="eeg")`. What should the fit do when multiple channel types are provided?

I'm guessing it should ignore non data channels (EOG, ECG, ...), and then what? What about `eeg`, `mag`, `grad`, `ieeg`, ...?

vferat:
For the moment, the algorithms have been written for EEG data only.

I propose one of these two solutions:
- to enforce `picks='eeg'` and delete the `picks` parameters from the public API

- enforce `picks` to accept only one data type at a time (impossible to have `picks=['eeg', 'meg']` or `picks=['MEG001", "Fz"]`)

mscheltienne:
Option (2) sounds good, especially as it's already tested + we keep the door open for expanding the use to fit multiple data types.

mscheltienne:
And probably easy to do, just a check on the shape of `.get_channel_types(unique=True)`. 
We can also pre-pick the data channels (at least to automatically ignore the EOG, ECG, ... channels and then check the shape of `.get_channel_types(unique=True, only_data_chs=True)`.

mscheltienne:
Which leads to rephrasing (2) as enforcing only 1 type of data channel in the instance to fit.

vferat:
Here again, 2 options:

- force user to provide an instance with `len(inst.get_channel_types(unique=True) == 1` and delete the `picks` parameter

- allow user to provide an instance with `len(inst.get_channel_types(unique=True) > 1` and enforce `picks` to only 1 type of data

The first is less user-friendly, the second may need more work

mscheltienne:
I think I would:

1. Edit the docstring of https://github.com/vferat/pycrostates/blob/31dcd72de70184ada12c218c802097cfe692a686/pycrostates/cluster/_base.py#L173-L194 to mention that only one type of data channels should be used for fitting. If `picks` selects multiple types of data channels, it will raise.

2. Do the check here https://github.com/vferat/pycrostates/blob/31dcd72de70184ada12c218c802097cfe692a686/pycrostates/cluster/_base.py#L209-L230 and make sure that only one type of data channel is present. If not, `raise RuntimeError` with a clear message.

WDYT?

mscheltienne:
And the check would be to make sure that the `picks` argument provided by the user does lead to one single type of data channel. So if we take the dataset above:

```
raw.pick_types(meg=True, eeg=True, eog=True)

ModK.fit(raw, picks=None)  # fails -> MEG + EEG
ModK.fit(raw, picks="meg")  # fails -> grad + mag
ModK.fit(raw, picks="eeg")  # works
ModK.fit(raw, picks="mag")  # works
ModK.fit(raw, picks="eog")  # fails, not a data channel

raw.pick_types(eeg=True, eog=True)

ModK.fit(raw, picks=None)  # works because eog is not a data channel
ModK.fit(raw, picks="data")  # works because only one type, eeg
```

pycrostates/cluster/_base.py
```diff
@@ -168,12 +168,28 @@ def _check_fit(self):
         assert self._fitted_data is not None
         assert self._labels_ is not None
 
+    def _check_picks(self, inst, picks):
+        inst_ = inst.copy()
+        inst_.pick(picks)
+        if len(inst_.get_channel_types(unique=True)) != 1:
+            ch_types = inst_.get_channel_types(unique=False)
+            ch_types, counts = np.unique(ch_types, return_counts=True)
```
mscheltienne, line 176:
```suggestion
    @staticmethod
    def _check_picks(info, picks):
        info = pick_info(info, picks, copy=True)
        if len(info.get_channel_types(unique=True)) != 1:
            ch_types = info.get_channel_types(unique=False)
            ch_types, counts = np.unique(ch_types, return_counts=True)
```

Let's work on the info instead of copying an entire data instance.

Also, this method does not interact with anything else than its argument (thus is static) and is only used in this class. If it was used elsewhere, I would suggest moving it as a separate function to `utils/_checks.py`, but as it's not, let's at least make it a `staticmethod`.

pycrostates/cluster/_base.py
```diff
@@ -207,6 +223,7 @@ def fit(
         info = inst.info
 
         # picks
+        self._check_picks(inst, picks)
```
mscheltienne, line 226:
```suggestion
        _BaseCluster._check_picks(inst.info, picks)
```

Syntax for `staticmethod`.

pycrostates/cluster/tests/test_kmeans.py
```diff
@@ -799,6 +799,25 @@ def test_fit_with_bads(caplog):
...
+        max_iter=100,
+        tol=1e-4,
+        random_state=1,
+    )
+    with pytest.raises(ValueError):
+        ModK_.fit(raw, picks=None)  # fails -> MEG + EEG
+    with pytest.raises(ValueError):
+        ModK_.fit(raw, picks="meg")  # fails -> grad + mag
+    with pytest.raises(ValueError):
+        ModK_.fit(raw, picks="data")  # fails -> eeg + grad + mag
```
mscheltienne, line 816: Could you add `match="matching string"`, with maybe part of the string which is generated based on the channels?

pycrostates/utils/_docs.py
```diff
@@ -117,6 +117,18 @@
...
+    "picks_fit"
+] = """
+picks : str | list | slice | None
+    Channels to include. Slices and lists of integers will be interpreted
+    as channel indices. Note that all channels must have the same type.
+    In lists, channel name strings (e.g., ['Fp1', 'Fp2']
+    will pick the given channels. "eeg" (default) will pick all eeg channels.
+    Note that channels in info['bads'] will be included if their
+    names or indices are explicitly provided."""
+
```
mscheltienne, line 131: This docstring is not used anywhere else than `_BaseCluster.fit`, so let's move it back to the function docstring directly.

pycrostates/io/ch_data.py
```diff
@@ -115,6 +115,34 @@ def get_data(self, picks=None) -> NDArray[float]:
...
+
+        picks = _picks_to_idx(self._info, picks, exclude=exclude)
+        data = self._data.copy()
+        data = data[picks, :]
+
+        info = pick_info(self._info, picks, copy=False)
+
+        self._data = data
+        self._info = info if isinstance(info, ChInfo) else ChInfo(info)
+        return self
```
mscheltienne, line 144:
```suggestion
        picks = _picks_to_idx(self._info, picks, exclude=exclude)
        self._info = pick_info(self._info, picks, copy=False)
        self._data = data[picks, :]
        return self
```

Simpler: `self._info` is always a `ChInfo`, and calling `pick_info` on a `ChInfo` returns a `ChInfo`.
With the change I suggest to `_check_picks`, this method is not used anymore, but it's a nice addition.

pycrostates/io/tests/test_ch_data.py
```diff
@@ -74,6 +78,28 @@ def test_ChData():
...
+        ("stim", [], ["STIM01"]),
+        ("data", [], ["Fpz", "Cz", "MEG01", "GRAD01"]),
+        ("all", [], ["Fpz", "Cz", "MEG01", "STIM01", "GRAD01", "EOG"]),
+        ("eeg", ["Fpz"], ["Cz"]),
+        ("data", ["Fpz"], ["Cz", "MEG01", "GRAD01"]),
+        ("all", ["Fpz"], ["Cz", "MEG01", "STIM01", "GRAD01", "EOG"]),
+    ],
+)
+def test_ChData_picks(picks, exclude, ch_names):
+    # test picks
```
mscheltienne, line 96:
```suggestion
def test_ChData_picks(picks, exclude, ch_names):
    """Test pick method."""
```

mscheltienne (commented):
A couple of changes I suggest, especially avoiding the copy of the data instance. Besides that, it looks good, and you are right not to add the auto-pick of data channels as I suggested, no need to add complexity here.

vferat:
### Fit

There is some issues when using `picks` and **bad channels** in `_BaseCluster.fit()`.

- if `picks` is a list of channels (e.g `picks=['Fp1', 'Fp2']`) and `inst.info['bads'] = ['Fp1']` then we want to use:
` _picks_to_idx(info, picks, none="all", exclude=[])`

- if `picks` is a channel types (e.g `picks='eeg'`) and `inst.info['bads'] = ['Fp1']` then we want to use:
` _picks_to_idx(info, picks, none="all", exclude='bads')`

and there is more `picks` / `bad` combinations that may cause troubles..

The solution I propose is to always keep bads channels and add a warning whenever a bad channel is used for fitting.

```
picks_ =  _picks_to_idx(info, picks, none="all", exclude=[])
_BaseCluster._check_picks(inst.info, picks_ )

info_= pick_info(info, picks_ )
if len(info_['bads']:
    warn(...)
```

### Predict

The current behaviour is of `_BaseCluster.predict()` is also problematic as we are rejecting **bads channels** by default, that may impact the spatial correlation (can we compare a segmentation predicted from 19 good channels (5 bads ignored) to a one from 24 good channels ?).

With this new solution, predicted segmentations from a `_BaseCluster` will always use the same number of channels ( = `len(_BaseCluster._info`) and:
- will raise if there is missing channels in the predict instance (`'Fp1'` in `_BaseCluster.info['ch_names']` but not in `inst.info['ch_names']`)
- will pick only subset of channels is there is more channels in the predict instance. (`'Fp1'` in `inst.info['ch_names']` but not in   `_BaseCluster.info['ch_names']`)
- will _raise_ or _warn_ if there is bads channels used for prediction.   (`'Fp1'` in `inst.info['bads']`)

mscheltienne:
Have a look at this small change, does it fix it? 
Else I'd like to look at that one a bit more.

mscheltienne:
Also, I just changed a bit the error message to avoid

image

mscheltienne:
Looking a bit more into it ->

> There is some issues when using `picks` and **bad channels** in `_BaseCluster.fit()`.
> 
> - if `picks` is a list of channels (e.g `picks=['Fp1', 'Fp2']`) and `inst.info['bads'] = ['Fp1']` then we want to use:
> ` _picks_to_idx(info, picks, none="all", exclude=[])`
> 
> - if `picks` is a channel types (e.g `picks='eeg'`) and `inst.info['bads'] = ['Fp1']` then we want to use:
> ` _picks_to_idx(info, picks, none="all", exclude='bads')`

`_picks_to_idx` handles this for us. When `exclude='bads'` is set, it will exclude bad channels except if they are explicitly provided.

```
import numpy as np
from mne import create_info
from mne.io import RawArray
from mne.io.pick import _picks_to_idx


info = create_info(["Fp1", "Fp2"], sfreq=1024, ch_types="eeg")
data = np.random.randn(2, 1024 * 10)
raw = RawArray(data, info)

_picks_to_idx(raw.info, "eeg", none="all", exclude="bads")
# >>> array([0, 1])

raw.info["bads"] = ["Fp1"]
_picks_to_idx(raw.info, ["Fp1", "Fp2"], none="all", exclude="bads")
# >>> array([0, 1])
```

So in the `.fit` code:

https://github.com/vferat/pycrostates/blob/31dcd72de70184ada12c218c802097cfe692a686/pycrostates/cluster/_base.py#L209-L230

This part checks for channels that are set as bads and warns users that they will not be used in the fit. 

https://github.com/vferat/pycrostates/blob/31dcd72de70184ada12c218c802097cfe692a686/pycrostates/cluster/_base.py#L238

This line retrieves the data array, excluding the `'bads'` (because `picks` is now the array of idx converted by `_picks_to_idx` in the code-block above).

https://github.com/vferat/pycrostates/blob/31dcd72de70184ada12c218c802097cfe692a686/pycrostates/cluster/_base.py#L245

This line creates the channel info and keeps the `'bads'` in `ModK.info['bads']`.

mscheltienne:
And for `predict`, you have this code block https://github.com/vferat/pycrostates/blob/31dcd72de70184ada12c218c802097cfe692a686/pycrostates/cluster/_base.py#L589-L620
which makes sure that we do not predict based on channels that were not used during the fit.

vferat:
### Fit
It doesn't really make sense to have differents picks for[ data (`picks`)](https://github.com/vferat/pycrostates/pull/66/files#diff-2680322096b53b1d5d6ff124ee69629b4e21d07ca45c26d0ed4664038c4226acL238) and[ info (`picks_bads_inc `)](https://github.com/vferat/pycrostates/pull/66/files#diff-2680322096b53b1d5d6ff124ee69629b4e21d07ca45c26d0ed4664038c4226acL245) as there is no need to keep track of bad channels not used for cluster computation. This is especially relevant while [comparing infos ](https://github.com/vferat/pycrostates/pull/66/files#diff-2680322096b53b1d5d6ff124ee69629b4e21d07ca45c26d0ed4664038c4226acL588)during prediction.

The only case were `BaseCluster._info["bads"] != []` is when user explicitly provide bad channel(s) with the `picks` parameter.

### Prediction
By default (and assuming the previous point) prediction should only be possible if `modK.info['ch_names']` is included in `inst.info['ch_names']` and none of them are in  `inst.info["bads"]` 


If user wants to predict using bad channels, it should be explicitly asked.
The `picks` parameter should be removed in favour of `include_bads: bool = False`.

In the case that some channel(s) are both in `inst.info["bads"]`  and  `modK.info['ch_names']` :
- if `include_bads = False` -> raise
- if `include_bads = True` -> warn and predict using the bad channel(s)


predict should use  channels from `BaseCluster._info["bads"]` as the were explicitly kept.

mscheltienne:
I just ran again an example in debug mode. In my mind it works, so I am probably missing something. Are you at CB tomorrow? Whiteboard required 🙏 

----------------------------------------

Here is the detailed example I just ran. I agree it's heavy machinery around this picking stuff, the point was to accommodate fitting on one recording and predicting on another, even if they do not have the same bad channels. The only requirement is that they must have an overall common set of channels (i.e. recorded on the same system), but not necessarily the same set of bad channels. When predicting, it selects the subset of common good channels on both the prediction data and on the `cluster_centers_`.

```
import numpy as np
from mne import create_info
from mne.io import RawArray

from pycrostates.cluster import ModKMeans


info = create_info(["Fp1", "Fp2", "CP2", "CP1"], sfreq=1024, ch_types="eeg")
info.set_montage("standard_1020")

# -- first raw recording -----------------------------------------------------
data = np.random.randn(4, 1024 * 10)
raw = RawArray(data, info)

# Fp2 set as bads
raw.info["bads"] = ["Fp2"]
ModK = ModKMeans(n_clusters=4, n_init=10, max_iter=100, tol=1e-4)
ModK.fit(raw)

"""
[_base.fit] WARNING: Channel Fp2 is set as bad and ignored. 
To include it, either remove it from 'inst.info['bads'] or provide its name 
explicitly in the 'picks' argument.
"""

# -- second raw recording -----------------------------------------------------
data = np.random.randn(4, 1024 * 10)
raw = RawArray(data, info)

# CP2 set as bad, Fp2 set as good
raw.info["bads"] = ["CP2"]
segmentation = ModK.predict(raw)

"""
[_base.predict] WARNING: Picked channel Fp2 was set as bads during fitting and 
will be ignored.
"""

# ModK shapes:
# >>> ModK.fitted_data.shape
# Out: (3, 10240)

# >>> ModK.cluster_centers_.shape
# Out: (4, 3)  # (4 clusters, 3 channels)


# When calling ModK.predict(raw) we get:
# --------------------------------------

# picks = None  -> those are the picks we select on the predicted instance, 
# i.e. raw with CP2 as bad.
# N.B: Maybe we should set the default to 'eeg' as well?

# picks_ = _picks_to_idx(inst.info, picks, none="all", exclude="bads")
# -> array([0, 1, 3]) which corresponds to Fp1, Fp2 and CP1. Make sense since
# CP2 is a bad channel.

# ch_ = ["Fp2"]  -> because Fp2 was set as bads on the recording used for
# fitting this ModK. 
# -> issues a warning to mention that Fp2 will be ignored during the prediction

# picks_ = [
#     ch
#     for k, ch in enumerate(inst.info["ch_names"])
#     if k in picks_ and ch not in self._info["bads"]
# ]
# -> ['Fp1', 'CP1'], same as picks before, but without Fp2 because it was a bad
# channel during fitting.

# picks_data = _picks_to_idx(inst.info, picks_, none="all", exclude=[])
# array([0, 3]) -> Again ["Fp1", "CP1"], just a conversion to IDX

# good_channels -> ['Fp1', 'CP2', 'CP1']
# The good channels used for fitting.

# picks_cluster_centers -> array([0, 2])
# The channels that were good during fitting and that are still good in the
# recording on which we predict.

# Note in good_channels and picks_cluster_centers, the indexing has to match
# the shape of ModK.cluster_centers_, i.e. 3 channels only. Thus array([0, 2])
# selects "Fp1" and "CP1" because:
#   0 -> Fp1
#   1 -> CP2 (since Fp2) was not part of the channel used for fitting
#   2 -> CP1

# Then we have in _predict_raw or _predict_epochs:
# ------------------------------------------------

# data = raw.get_data(picks=picks_data)
# -> picks = (0, 3) which are Fp1, CP1

# cluster_centers_ = deepcopy(self._cluster_centers_)
# cluster_centers_ = cluster_centers_[:, picks_cluster_centers]
# -> picks_cluster_centers = (0, 2) which are also Fp1 and CP1
```

mscheltienne:
Also, we should probably call `_check_picks` in `predict` as well because it should also be limited to a single data type.

mscheltienne:
@vferat 🤯 🤯 

```
import numpy as np
from mne import create_info
from mne.io import RawArray

from pycrostates.cluster import ModKMeans


info_ = create_info(["Fp1", "Fp2", "CP2", "CP1"], sfreq=1024, ch_types="eeg")
info_.set_montage("standard_1020")
data = np.random.randn(4, 1024 * 10)

ModK_ = ModKMeans(n_clusters=4, n_init=10, max_iter=100, tol=1e-4)


# ---- 1 channel bad ----
info = info_.copy()
info['bads'] = ["Fp2"]
raw = RawArray(data, info)

ModK = ModK_.copy()
ModK.fit(raw, picks="eeg")
# -> Select all but the bad channel Fp2
# -> Warning that Fp2 is not used
# -> Remove Fp2 from ModK.info


# overall, same channels in the instance used for prediction
info = info_.copy()
raw_predict = RawArray(data, info)
ModK.predict(raw_predict, picks="eeg")  # -> works, with warning that Fp2 is missing from ModK.info
ModK.predict(raw_predict, picks=["Fp1", "CP2", "CP1"])  # -> works, same channels as ModK.info
raw_predict.info["bads"] = ["Fp2"]
ModK.predict(raw_predict, picks="eeg")  # -> works, same channels as ModK.info
raw_predict.info["bads"] = ["Fp1"]
ModK.predict(raw_predict, picks="eeg")  # -> fails, because ModK.info includes Fp1 which is bad in prediction instance
ModK.predict(raw_predict, picks=["Fp1", "CP2", "CP1"])  # works, with warning because you predict based on a bad channel Fp1
ModK.predict(raw_predict, picks=["CP2", "CP1"])  # -> fails, because ModK.info includes Fp1 which is missing from prediction instance selection


# one more channel in the instance used for prediction
info_ = create_info(["Fp1", "Fp2", "Fpz", "CP2", "CP1"], sfreq=1024, ch_types="eeg")
info_.set_montage("standard_1020")
data = np.random.randn(5, 1024 * 10)
raw_predict = RawArray(data, info)

ModK.predict(raw_predict, picks="eeg")  # -> works, with warning because Fpz, Fp2 are missing from ModK.info
ModK.predict(raw_predict, picks=["Fp2", "Fpz", "CP2", "CP1"])  # -> fails, because ModK.info includes Fp1 which is missing from prediction instance selection
ModK.predict(raw_predict, picks=["Fp1", "Fpz", "CP2", "CP1"])  # -> works, with warning because Fpz is missing from ModK.info


# ---- 1 channel bad but fit on it ----
info = info_.copy()
info['bads'] = ["Fp2"]
raw = RawArray(data, info)

ModK = ModK_.copy()
ModK.fit(raw, picks=["Fp1", "Fp2", "CP2", "CP1"])
# -> Select all channels
# -> Warning that Fp2 is used
# -> Set Fp2 as bads in ModK.info


# overall, same channels in the instance used for prediction
info = info_.copy()
raw_predict = RawArray(data, info)
ModK.predict(raw_predict, picks="eeg")  # -> works, with warning because a channel is bads in ModK.info
raw_predict.info["bads"] = []
ModK.predict(raw_predict, picks=["Fp1", "Fp2", "CP2", "CP1"])  # -> works, with warning because a channel is bads in ModK.info
ModK.predict(raw_predict, picks=["Fp1", "CP2", "CP1"])  # -> fails, because Fp2 is used in ModK.info
raw_predict.info["bads"] = ["Fp2"]
ModK.predict(raw_predict, picks="eeg")  # -> fails, because Fp2 is used in ModK.info
ModK.predict(raw_predict, picks=["Fp1", "Fp2", "CP2", "CP1"])  # -> works, because same channels as ModK.info
raw_predict.info["bads"] = ["Fp1"]
ModK.predict(raw_predict, picks="eeg")  # -> fails, because ModK.info includes Fp1 which is bad in prediction instance
ModK.predict(raw_predict, picks=["Fp1", "CP2", "CP1"])  # -> fails, because ModK.info includes bad Fp2
ModK.predict(raw_predict, picks=["Fp1", "Fp2", "CP2", "CP1"])  # -> works, because same channels as ModK.info (with 2 warning for Fp1 and Fp2)
ModK.predict(raw_predict, picks=["CP2", "CP1"])  # -> fails, because ModK.info includes Fp1, Fp2
```

mscheltienne:
Let me know when you want a review, I'll keep myself from looking at this mess of channel selection until then 😛

vferat:
@mscheltienne you can have a look at it.
IdK if it's worth spending more time on the error/warning messages. There is a lot of different cases with the pick parameters which can be hard to resolve

mscheltienne (approved):
LGTM, I improved the error messages with full sentences. I checked all the logic and had a small headache with the `set()` comparison in `predict()`, but it works great!

Also, I added in `_base` a `_check_unfitted` which now raises if you try to fit a ModKMeans that is already fitted. 2 reasons behind this addition:
- don't overwrite by mistake a ModKMeans. The error message mentions that you can do `ModK.fitted = False` to unfit a cluster algorithm
- `ModK.fit(raw1)` followed by `ModK.fit(raw2)` with different channels.. was silently not working anyway

mscheltienne:
I'll have a look at the errors. The doc build is simple, but the issue with `mne-main`.. no clue for now.
It looks like it's topomap-related, and I know they changed how the interpolation for topomap is done in MNE 1.1. Probably related.

mscheltienne:
The tutorial failing is https://github.com/vferat/pycrostates/blob/main/tutorials/group_level_analysis/plot_group_level_with_individual_clusters.py

and indeed I don't get what the fit is doing here https://github.com/vferat/pycrostates/blob/eef67e7906a619cf5960431cafc2f35161e2f765/tutorials/group_level_analysis/plot_group_level_with_individual_clusters.py#L57

@vferat Can you have a look at that one and fix the logic behind?

vferat:
The tutorial was failing because we were fitting the same Kmeans several times. It should be fixed now ( we now create a new Kmeans instance for each subject).

Error messages and warning are clearer now, thanks a lot !

I'm not sure that the `data` and `all`  **pick options** are relevant in our case, as they should always correspond to a data type (i.e `eeg`). However, it is simpler to maintain this way and should not cause trouble, so I'm ok keeping them.

mscheltienne:
Yes, I could see why it was failing. My point is why are we even fitting a `ModK` at every iteration of the loop? You don't use anything from this `ModK` after fitting it, and you add `gfp_peaks.get_data()` to the list which stores results from that loop. Am I missing something?

Or based on the variable name `individual_cluster_centers`, I guess you are not appending the correct element to it?

mscheltienne (approved):

mscheltienne:
For `data` and `all`, I agree they don't make a lot of sense for `pycrostates`. Maybe we could at least remove them from the docstrings?

vferat:
> For `data` and `all`, I agree they don't make a lot of sense for `pycrostates`. Maybe we could at least remove them from the docstrings?

I would rather keep the description in the docstring ifvwe allow users to use them ( even if they are useless). The other solution would be not to allow the use of `'all'` and `'data'`

mscheltienne:
That would be a pain to do, so let's keep it as is.

PR #73, opened 2022-09-05
Add _check_picks_uniqueness for extract_gfp_peaks

vferat:
Check for picks uniqueness in `extract_gfp_peaks`.  It is wrong to compute GFP (np.std) across different channel types.

We can have to approch for the implementation
- compute GFP peaks on picked channels and return all channels (even the not picked ones).
- compute GFP peaks on picked channels and return only picked channels.

I don't see any use case of the first one, but don't have any reason not to do so...

 @mscheltienne any hint ?

### DIFF
--- a/pycrostates/preprocessing/extract_gfp_peaks.py
+++ b/pycrostates/preprocessing/extract_gfp_peaks.py
@@ -86,7 +86,9 @@ def extract_gfp_peaks(
 
     # retrieve picks
     picks = _picks_to_idx(inst.info, picks, none="all", exclude="bads")
-    picks_all = _picks_to_idx(inst.info, inst.ch_names, none="all", exclude="bads")
+    picks_all = _picks_to_idx(
+        inst.info, inst.ch_names, none="all", exclude="bads"
+    )
     _check_picks_uniqueness(inst.info, picks)
     # retrieve data array
     kwargs = (
@@ -114,7 +116,7 @@ def extract_gfp_peaks(
             ind_peaks = _extract_gfp_peaks(data[k, :, :], min_peak_distance)
             print(ind_peaks)
             if return_all:
-                peaks.append(data_all[k, :, ind_peaks].T) # TODO: why .T
+                peaks.append(data_all[k, :, ind_peaks].T)  # TODO: why .T
             else:
                 peaks.append(data[k, :, ind_peaks].T)
         peaks = np.hstack(peaks)
--- a/pycrostates/preprocessing/extract_gfp_peaks.py
+++ b/pycrostates/preprocessing/extract_gfp_peaks.py
@@ -51,7 +51,7 @@ def extract_gfp_peaks(
         Note that channels in ``info['bads']`` will be included if their
         names or indices are explicitly provided.
     return_all : bool
-        If True, the returned `~pycrostates.io.ChData` instance will 
+        If True, the returned `~pycrostates.io.ChData` instance will
         include all channels.
         If False (default), the returned `~pycrostates.io.ChData` instance will
         only include channels used for GFP computation (i.e ``picks``).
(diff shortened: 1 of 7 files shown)


pycrostates/utils/_checks.py
```diff
@@ -279,3 +280,18 @@ def _check_tmin_tmax(inst, tmin, tmax):
             f"{inst.times[-1]}s instance."
         )
     return tmin, tmax
+
+
+def _check_picks_uniqueness(info, picks):
+    info = pick_info(info, picks, copy=True)
```
mscheltienne, line 286:
```suggestion
def _check_picks_uniqueness(info, picks):
    """Check that the provided picks yield a single channel type."""
    info = pick_info(info, picks, copy=True)
```

mscheltienne:
I don't get your 2 options, especially:

> compute GFP peaks on picked channels and return all channels (even the not picked ones).

Based on the code in this PR, it will:
- raise if you provides picks that include multiple channel types
- return a ChData with only the picked channels

Sounds reasonable to me.

mscheltienne:
But then what is the data for this ECG channel? That sounds very confusing as you would get GFP peaks for the EEG channels, and a regular channel for the ECG?

vferat:
> But then what is the data for this ECG channel? That sounds very confusing as you would get GFP peaks for the EEG channels, and a regular channel for the ECG?

`ChData `contains samples at timepoints where GFP peaks occur. We can compute when GFP peaks occur on some channels ( an force only one channel type i.e EEG),  but still extract all data ( all channel type i.e EEG + ECG) from the corresponding samples

pycrostates/preprocessing/extract_gfp_peaks.py
```diff
@@ -40,14 +41,20 @@ def extract_gfp_peaks(
...
         ``['Fp1', 'Fp2']``) will pick the given channels. Can also be the
         string values “all” to pick all channels, or “data” to pick data
         channels. ``"eeg"`` (default) will pick all eeg channels.
         Note that channels in ``info['bads']`` will be included if their
         names or indices are explicitly provided.
+    return_all : bool
+        If True, output ChData instance will include all channels.
+        If False, output ChData instance will only include channels
+        used for GFP computation (i.e picks).
+        Default to False.
```
mscheltienne, line 57:
```suggestion
    return_all : bool
        If True, the returned `~pycrostates.io.ChData` instance will 
        include all channels.
        If False (default), the returned `~pycrostates.io.ChData` instance will
        only include channels used for GFP computation (i.e ``picks``).
```

pycrostates/preprocessing/extract_gfp_peaks.py
```diff
@@ -70,23 +86,39 @@ def extract_gfp_peaks(
...
     kwargs = (
         dict()
         if isinstance(inst, BaseEpochs)
         else dict(reject_by_annotation=reject_by_annotation)
     )
     data = inst.get_data(picks=picks, tmin=tmin, tmax=tmax, **kwargs)
+    if return_all:
+        data_all = inst.get_data(
+            picks=picks_all, tmin=tmin, tmax=tmax, **kwargs
+        )
```
mscheltienne, line 103:
```suggestion
    data = inst.get_data(picks=picks_all if return_all else picks, tmin=tmin, tmax=tmax, **kwargs)
```

let's call get_data only once

mscheltienne:
I'll review later or tomorrow, I was starting to propose dumb changes ;)

mscheltienne:
I made a change, I moved the retrival of the data array with `inst.get_data()` to where we need the data array, and I made sure to select only what was needed (e.g. for epochs, doing `inst[k].get_data()` returns an array of shape `(1, n_channels, n_samples)`.

The overall goal was to reduce the memory required by this function to run by avoiding calling twice `.get_data()` on `picks` and `picks_all`.

mscheltienne (approved):
Good to merge?

PR #77, opened 2022-09-09
Compare projectors if MNE is above 1.2, fix matplotlib deprecation and drop support for Python 3.7

### LINKED ISSUE #75: [ENH] Compare projectors when comparing ChInfo/Info objects
Associated bullet point: https://github.com/vferat/pycrostates/blob/2d9a3fd4ddd13f1a962f3a83b79cafdc00c87061/pycrostates/io/meas_info.py#L378

There are still 2 attributes that are not compared when doing `chinfo1 == chinfo2`. For the compensation grade, it's a MEG feature, and for now I did not check how it looks like.
`projs` are however a MEG and EEG feature, so we should add a comparison. The easy way is to wait for https://github.com/mne-tools/mne-python/issues/11137 to be solved, and then compare projectors for MNE version >= 1.2 (if it's added in 1.2). 
To compare versions, we can use https://github.com/mne-tools/mne-python/blob/879829e3d4da228f0c022d8f1ca4abf6e4599c81/mne/utils/check.py#L68-L119

### LINKED ISSUE #98: [MNT] Matplotlib deprecation warning
Azure is showing:

```
[warning]The get_cmap function will be deprecated in a future version. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
```

To keep in mind and fix.

mscheltienne:
Closes #75 
Closes #98


vferat (approved):
Looks great, nothing to add

PR #79, opened 2022-09-09
Use invert=True instead of custom code in _annotations_starts_stops

### LINKED ISSUE #78: [BUG] Overlapping 'bad' annotations cause wrong BaseCluster._segment_raw() behaviour
The `reject_by_annotations `argument is not working properly in `BaseCluster._segment_raw()`

```
import os
import numpy as np
import mne

sample_data_folder = mne.datasets.sample.data_path()
sample_data_raw_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file)
raw.set_annotations(None)
    
annotation_0 = mne.Annotations(onset=100, duration=100, description='bad')
annotation_1 =  mne.Annotations(onset=130, duration=30, description='bad')

raw_01 = raw.copy()
raw_01.set_annotations( annotation_0 + annotation_1) #overlapping 'bad' annotations

```

We apply the logic from `_segment_raw()` found [here ](https://github.com/vferat/pycrostates/blob/1cb8d0c27e48ce9e6aa0c57ea2773dc89bd22d42/pycrostates/cluster/_base.py#L811)

```
onsets, ends = _annotations_starts_stops(raw_01, ["BAD"])
onsets = onsets.tolist() + [raw_01.get_data().shape[-1] - 1]
ends = [0] + ends.tolist()

for onset, end in zip(onsets, ends):
    print(end, onset)
```
output:
```
0 60061
120123 78080
96098 166799
```

The segments are completely messed up due to the overlapping 'bad' annotations. 
The expect behaviour would be :
```
0 60061 # start of recording - start of first 'bad' annotation
120123 166799 # end of first 'bad' annotation - end of recording
```

No sure if we should fix the in Pycrostates or if it should be fix in `_annotations_starts_stops` in MNE

```
mne.__version__
1.1.0
```

vferat:
Closes #78

### DIFF
--- a/pycrostates/cluster/_base.py
+++ b/pycrostates/cluster/_base.py
@@ -810,24 +810,21 @@ class _BaseCluster(ABC, ChannelsMixin, ContainsMixin, MontageMixin):
 
         if reject_by_annotation:
             # retrieve onsets/ends for BAD annotations
-            onsets, ends = _annotations_starts_stops(raw, ["BAD"])
-            onsets = onsets.tolist() + [data.shape[-1] - 1]
-            ends = [0] + ends.tolist()
-
+            onsets, ends = _annotations_starts_stops(raw, ["BAD"], invert=True)
             segmentation = np.full(data.shape[-1], -1)
 
             for onset, end in zip(onsets, ends):
                 # small segments can't be smoothed
-                if factor != 0 and onset - end < 2 * half_window_size + 1:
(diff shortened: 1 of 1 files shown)


mscheltienne (approved):
Tested a bit on a simple example, LGTM.

PR #87, opened 2022-10-10
[DOC] Improve documentation and module docstrings

mscheltienne:
3 changes:
- I removed the duplicate title on the `index` page
- I re-ordered a bit the installation page, mentioning first the version of Python required and then the dependencies
- I improved a bit the docstrings of the different modules which appear on the API Reference page

For this last point, the idea was to detail a bit the general structure of the module, e.g. when all functions/classes within this module adhere to a similar design.


vferat:
Thanks @mscheltienne, the is a very nice update. I really like the new docstrings of the different modules that give a better understanding of the structure of the library.

PR #91, opened 2022-10-27
Add transition probabilities

vferat:
Add support to compute microstates transition probabilities.

## Improvments
- [x] Add `_compute_transition_probabilities` method to `_BaseSegmentation`.
- [x] Add test for `_compute_transition_probabilities`.
- [x] Add tutorial for `_compute_transition_probabilities`.

## Fixes
- [x] Fix gfp normalization to exclude unlabeled timepoints in `_BaseSegmentation._compute_microstate_parameters `.


vferat:
Hey @mscheltienne,

Please have a look. I added the `pycrostates.segmentation.RawSegmentation.compute_transition_probabilities` method which computes transitions from one state to another.

I also removed the segmentation part of the `plot_modK `tutorial and created `plot_segmentation `a dedicated tutorial.

With this PR, the `docs `dependencies requires `seaborn`. Let me know if you think I should add a note in tutorials using it.

We can decide later if we want to also include it in the main dependencies and create dedicated viz functions with it.

tutorials/clustering/plot_segmentation.py
```diff
@@ -0,0 +1,205 @@
+"""
+Microstate Segmentation
+=======================
+
+This tutorial introduces the concept while working with microstate
+segmentations.
```
mscheltienne, line 6:
```suggestion
This tutorial introduces the concept of segmentation.
```

tutorials/clustering/plot_segmentation.py
```diff
@@ -0,0 +1,205 @@
...
+"""
+
+#%%
+# .. include:: ../../../../links.inc
+
+#%%
+# Here we fit a modified K-means with a sample dataset.
+# For more details about the fitting procedure,
+# please refer to :ref:`this <fit>`
+# tutorial.
```
mscheltienne, line 16:
```suggestion
# please refer to :ref:`this tutorial <fit>`.
```

tutorials/clustering/plot_segmentation.py
```diff
@@ -0,0 +1,205 @@
...
+n_clusters = 5
+ModK = ModKMeans(n_clusters=n_clusters, random_state=42)
+
+ModK.fit(raw, n_jobs=5)
+ModK.reorder_clusters(order=[4, 1, 3, 2, 0])
+ModK.rename_clusters(new_names=['A', 'B', 'C', 'D', 'F'])
+ModK.plot()
+
+#%%
+# Once a set a cluster centers has been computed, we can use it  to
```
mscheltienne, line 38:
```suggestion
# Once a set of cluster centers has been computed, we can use it  to
```

tutorials/clustering/plot_segmentation.py
```diff
@@ -0,0 +1,205 @@
...
+# predict the microstate segmentation using the :meth:`pycrostates.cluster.ModKMeans.predict`
+# method. This method returns either a `~pycrostates.segmentation.RawSegmentation` or
+# `~pycrostates.segmentation.EpochsSegmentation` depending on the input instance.
+segmentation = ModK.predict(raw, reject_by_annotation=True, factor=10,
+                            half_window_size=10, min_segment_length=5,
+                            reject_edges=True)
+
+#%%
+# The ``factor`` and ``half_window_size`` allows control of the smoothing algorithm introduced
```
mscheltienne, line 47:
```suggestion
# The ``factor`` and ``half_window_size`` control the smoothing algorithm introduced
```

mscheltienne (commented):
I'll have to run the tutorial once, but otherwise LGTM.

pycrostates/segmentation/segmentation.py
```diff
@@ -147,6 +156,87 @@ def compute_parameters(
...
+        Parameters
+        ----------
+        stat : str
+            Aggregate statistic to compute each transition. Can be:
+
+            * ``count``: show the number of observations of each transition.
+            * ``probability`` or ``proportion``: normalize such probabilities
+                along the first axis is always equal to 1.
+            * ``percent``: normalize such probabilities along the first axis
+                is always equal to 100.
```
mscheltienne, line 173:
```suggestion
            - ``count``: show the number of observations of each transition.
            - ``probability`` or ``proportion``: normalize such probabilities
                along the first axis is always equal to 1.
            - ``percent``: normalize such probabilities along the first axis
                is always equal to 100.
```

Still not good, let's try with a different bullet point style since stars are also used for bold/italic/...

mscheltienne:
I had a quick look at the doc build, on the API page. At the moment, it's not clear what is the difference between `get_expected_transition_matrix` and `get_transition_matrix`. They both have the same description.

vferat:
@mscheltienne please have a look ! 

I change the prefix `get_` to `compute_` in methods as it is not that straightforward.

I can't get satisfied with display of `stat` options in `compute_transition_matrix ` and  `compute_expected_transition_matrix `([here](https://pycrostates--91.org.readthedocs.build/en/91/api/generated/pycrostates.segmentation.RawSegmentation.html#pycrostates.segmentation.RawSegmentation.compute_transition_matrix)). 

I leave it like that, I let you merge if it seems good to you

mscheltienne:
Hey, I did notice the ping for review, but I lack time this week to review this (end of the contract, trying to wrap things up which is impossible, but well..). Hopefully, I can review it next week!

tutorials/segmentation/00_segmentation.py
```diff
@@ -0,0 +1,300 @@
...
+ax.set_ylabel("From")
+ax.set_xlabel("To")
+
+#%%
+# The difference between the observed transition probability matrix
+# ``T_observed`` and the theoretical transition probability matrix
+# ``T_expected`` reveals particular dynamic present in the segmentation.
+# Here we observe that the transition from state ``B`` to state ``A`` appears
+# in larger proportion than expected while the transition from state ``B`` to
+# state ``C`` is less observed than expected.
```
mscheltienne, line 283: @vferat Please double-check this. Previously it was stated from B to C in both cases.

pycrostates/segmentation/segmentation.py
```diff
@@ -147,9 +145,187 @@ def compute_parameters(
...
+        T = np.zeros(shape=(n_clusters, n_clusters))
+        # number of transitions
+        for i, j in zip(labels, labels[1:]):
+            if i == -1 or j == -1:
+                continue  # ignore unlabeled
+            T[i, j] += 1
+
+        # transform to probability
+        if stat != "count":
+            T = T / T.sum(axis=1, keepdims=True)
```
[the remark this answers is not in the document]
mscheltienne: For now the tests are failing because of this change since you included a test that covers this case.
Alternatively, to detecting and handling the issue at the beginning, let's at least use a vectorized version of the oop:

```
with np.errstate(divide='ignore'):
    T = T / T.sum(axis=1, keepdims=True)
    np.nan_to_num(T, posinf=0, copy=False)
```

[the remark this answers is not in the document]
mscheltienne: Ok, let's keep the handing at the end then. I was also thinking about the opposite case where you have more labels than `n_clusters` (which could have been handled together at the beginning of the function), but I guess this case should not occur without tempering into the private variables of a segmentation. 

I fixed the test with `np.errstate(divide='ignore', invalid="ignore"):`, because 5 / 0 yields inf while 0 / 0 yields nan in numpy, and that's 2 different errors.

pycrostates/segmentation/segmentation.py
```diff
@@ -147,9 +145,187 @@ def compute_parameters(
...
+
+        Parameters
+        ----------
+        stat : str
+            Aggregate statistic to compute transitions. Can be:
+
+            * ``probability`` or ``proportion``: normalize count such as the
+              probabilities along the first axis is always equal to ``1``.
+            * ``percent``: normalize count such as the
+              probabilities along the first axis is always equal to ``100``.
```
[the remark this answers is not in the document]
vferat: perfect !

vferat:
> Looks very good to me. The changes I made are:
> 
> * Rename the examples to the numerical system, 00_tutorial1, 10_tutorial2, ...
> * Improve the phrasing in the tutorial, fix some spelling tpyos, added notes about optional dependencies, and added subtitles.
> * Change how the docstring render
> * Minor optimization of `compute_transition_matrix`
> 
> Also, looking through the code, I don't see a reason why `compute_parameters`, `compute_transition_matrix`, and `compute_expected_transition_matrix` should be a one-line wrapper around a static method. IMO, the code from the static method should be moved back to the main method directly. WDYT?

I like the fact to be able to compute transition matrices without the need to create a `Segmentation` instance. I can be handy when you only have a sequence of labels, without original data and/or cluster centers.
Maintenance wise, I don't see any issue keeping the current implementation... Except if you haveany reason to do so ?

tutorials/segmentation/00_segmentation.py
vferat: all good

mscheltienne:
Very good point. But then, maybe let's move them as public documented low-level function, kind of like the `_array` function from MNE, e.g. `psd_array_welch`, ...
Currently, all 3 are private static methods, right?

vferat:
I agree, we can move theses static methods outside the `_BaseSegmentation` and create low-level functions.
In this case, we need to add more checks. The current implementation assumes that labels are ordered (i.e.  `np.unique(labels)` are in  `np.arange(n_clusters)`.  An example with `n_clusters = 4 ` but `np.unique(labels) = [0,1,2,6]` will give wrong results.

mscheltienne:
Yes, that was more or less the check I was thinking about instead of the `sum > 0` in the for loop 😅 
Do you want to give it a shot? Else I can do it tomorrow (I got time to kill this week ^^').

For the naming, maybe the transitions could be in `segmentation/transition.py`, and the `computer_parameters` could be in `segmentation/parameters.py`?

pycrostates/segmentation/segmentation.py
```diff
@@ -147,9 +145,184 @@ def compute_parameters(
...
+        # ignore transition to itself (i.e. 1 -> 1)
+        if ignore_self:
+            labels = [s for s, _ in itertools.groupby(labels)]
+
+        T = np.zeros(shape=(n_clusters, n_clusters))
+        # number of transitions
+        for i, j in zip(labels, labels[1:]):
+            if i == -1 or j == -1:
+                continue  # ignore unlabeled
+            T[i, j] += 1
```
mscheltienne, line 220: Also, there is probably a way to vectorize this one and gain a significant speed improvement, especially as `labels` can be a large array depending on the recording.

vferat:
Go on if you already an idea how to implement it

vferat:
`computer_parameters` needs data for Correlation and Global explained variance computations, so not sure it's worth creating a low level function for this one

mscheltienne:
Hmm yes, let's keep it as a method only then.

mscheltienne:
That should cover the general structure, let me know what you think.
I kept public/private function(s) to be able to skip the (not yet implemented) checks on `labels`/`n_clusters`. I'll add the checks on those 2 arguments tomorrow.

vferat:
the  general structure looks good to me, you can go on

pycrostates/segmentation/transitions.py
```diff
@@ -156,3 +156,55 @@ def _compute_expected_transition_matrix(
...
+            f"first state encoded as '0'. '{states}' are invalid."
+        )
+    # check if the states match the number of clusters
+    if len(states) != n_clusters:
+        raise ValueError(
+            "The argument 'labels' must contain the labels of each timepoint "
+            "encoded as consecutive positive integers (0-indexed), with the "
+            f"first state encoded as '0'. '{states}' are invalid with "
+            f"'n_clusters' set to '{n_clusters}'."
+        )
```
vferat, line 203: If would rather handle the case where one state never occur in the segmentation ( i.e. `labels = [1,2,3,4]` and` n_clusters=[5]`). Something like:

```python
if not np.isin(states, np.arange(n_clusters)):
    raise ...
```

pycrostates/segmentation/tests/test_segmentation.py
```diff
@@ -289,79 +285,6 @@ def test_invalid_segmentation(Segmentation, inst, bad_inst, caplog):
...
-        else np.unique(labels).size
-    )
-    t = _compute_transition_matrix(
-        labels, n_clusters=n_clusters, ignore_self=ignore_self
-    )
-    assert isinstance(T, np.ndarray)
-    assert t.shape == (n_clusters, n_clusters)
-    assert np.allclose(t, T)
-
-
```
vferat: Any reason to remove quantitative tests here ?

pycrostates/segmentation/tests/test_segmentation.py
```diff
@@ -389,42 +315,6 @@ def test_compute_transition_matrix_stat():
...
-            labels_, n_clusters, ignore_self=True, stat="probability"
-        )
-        Ts.append(T)
-    boostrap_T = np.array(Ts).mean(axis=0)
-    expected_T = _compute_expected_transition_matrix(
-        labels, n_clusters, ignore_self=True, stat="probability"
-    )
-    assert np.allclose(boostrap_T, expected_T, atol=1e-2)
-
-
```
vferat: Any reason to remove quantitative tests here ?

pycrostates/segmentation/tests/test_segmentation.py
mscheltienne: I moved the quantitative (great!) tests to `test_transitions.py` as they only use the private functions defined in `transitions.py`

mscheltienne: (I closed the duplicate comment below, feel free to re-open if I missed something)

pycrostates/segmentation/transitions.py
mscheltienne: I started with more constrained 'labels' and 'n_clusters', but indeed that should work nicely. I can't think of a breaking edge case. I changed the logic and the tests accordingly.

mscheltienne:
LGTM, have a look @vferat

pycrostates/segmentation/transitions.py
```diff
@@ -0,0 +1,197 @@
...
+            "representing unlabelled segments. Negative integers except -1 "
+            "are invalid."
+        )
+    # check that the labels are in the range of clusters
+    states = sorted([elt for elt in np.unique(labels) if 0 <= elt])
+    if not np.all(np.isin(states, np.arange(n_clusters))):
+        raise ValueError(
+            "The argument 'labels' must contain the labels of each timepoint "
+            "encoded as consecutive positive integers (0-indexed), between 0 "
+            f"and 'n_clusters' ({n_clusters}. '{states}' are invalid."
```
vferat, line 196: This error message seems a bit misleading: ` '{states}' are invalid.` will print both allowed and non-allowed states. I would rather print only non-allowed states are rephrase/delete the last part of the error message

mscheltienne: Done.

vferat (commented):
LGTM

PR #92, opened 2022-10-28
[ENH] Add support for AAHC clustering

rkobler:
Hi pycrostates team @vferat @mscheltienne ! 👋 

Thank you for working on a python repository that integrates microstate analysis with MNE python.

I recently started using the repository and felt like adding support for the atomize and agglomerate hierarchical clustering (AAHC) algorithm. The implementation is inspired by the matlab code provided in the eeglab plugin Microstates v 1.2 [1].

Please, let me know if you see merit in adding this method to the package.

Best,
Reinmar

[1] https://www.thomaskoenig.ch/Download/EEGLAB_Microstates/Microstates1.2.zip

### DIFF
--- a/pycrostates/cluster/__init__.py
+++ b/pycrostates/cluster/__init__.py
@@ -15,6 +15,6 @@ same channels) using the ``predict`` method. This method will return either a
 to segment."""
 
 from .kmeans import ModKMeans  # noqa: F401
-from .aahc import AAHCluster
+from .aahc import AAHCluster  # noqa: F401
 
 __all__ = ("ModKMeans", "AAHCluster")
--- a/pycrostates/cluster/__init__.py
+++ b/pycrostates/cluster/__init__.py
@@ -14,7 +14,7 @@ same channels) using the ``predict`` method. This method will return either a
 :class:`~pycrostates.segmentation.EpochsSegmentation` depending on the dataset
 to segment."""
 
-from .kmeans import ModKMeans  # noqa: F401
 from .aahc import AAHCluster  # noqa: F401
+from .kmeans import ModKMeans  # noqa: F401
 
 __all__ = ("ModKMeans", "AAHCluster")
--- a/pycrostates/cluster/__init__.py
+++ b/pycrostates/cluster/__init__.py
@@ -15,5 +15,6 @@ same channels) using the ``predict`` method. This method will return either a
 to segment."""
 
 from .kmeans import ModKMeans  # noqa: F401
+from .aahc import AAHCluster
 
-__all__ = ("ModKMeans",)
+__all__ = ("ModKMeans", "AAHCluster")
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -86,7 +86,7 @@ class AAHCluster(_BaseCluster):
                 f"{ch_count} {ch_type.upper()}"
                 for ch_type, ch_count in zip(ch_types, ch_counts)
             ]
-            GEV = int(self._GEV_ * 100)
+            GEV = f"{self._GEV_ * 100:.2f}"
         else:
             n_samples = None
             ch_repr = None
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -107,7 +107,6 @@ class AAHCluster(_BaseCluster):
 
     @copy_doc(_BaseCluster.__eq__)
     def __eq__(self, other: Any) -> bool:
-        """Equality == method."""
         if isinstance(other, AAHCluster):
             if not super().__eq__(other):
                 return False
@@ -131,7 +130,6 @@ class AAHCluster(_BaseCluster):
 
     @copy_doc(_BaseCluster.__ne__)
     def __ne__(self, other: Any) -> bool:
-        """Different != method."""
         return not self.__eq__(other)
 
     @copy_doc(_BaseCluster._check_fit)
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -303,8 +303,7 @@ class AAHCluster(_BaseCluster):
             Tolerance for convergence.
         v0 : numpy.ndarray or None (default)
             Initial estimate of the principal vector with dimensions
-            (channels, )
-            If None, a random initialization is used
+            (channels,). If None, a random initialization is used.
         max_iter : int
             Maximum number of iterations to estiamte the first principal
             component.
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -21,7 +21,6 @@ from ._base import _BaseCluster
 try:
     from numba import njit
 except ImportError:
-
     def njit(cache=False):
         def decorator(func):
             @func_wraps(func)
@@ -33,6 +32,7 @@ except ImportError:
         return decorator
 
 
+@fill_doc
 class AAHCluster(_BaseCluster):
     r"""Atomize and Agglomerate Hierarchical Clustering (AAHC) algorithm.
 
@@ -41,7 +41,6 @@ class AAHCluster(_BaseCluster):
     Parameters
     ----------
     %(n_clusters)s
-
     ignore_polarity : bool
         If true, polarity is ignored when computing distances.
     normalize_input : bool
@@ -81,7 +80,7 @@ class AAHCluster(_BaseCluster):
     def _repr_html_(self, caption=None):
         from ..html_templates import repr_templates_env
 
-        template = repr_templates_env.get_template("ModKMeans.html.jinja")
+        template = repr_templates_env.get_template("AAHCluster.html.jinja")
         if self.fitted:
             n_samples = self._fitted_data.shape[-1]
             ch_types, ch_counts = np.unique(
@@ -262,7 +261,7 @@ class AAHCluster(_BaseCluster):
                 members = assignment == c
                 if ignore_polarity:
                     v, _ = AAHCluster._first_principal_component(
-                        data[:, members], tol
+                        data[:, members], tol, cluster[:, c]
                     )
                     cluster[:, c] = v
                 else:
@@ -279,16 +278,45 @@ class AAHCluster(_BaseCluster):
     @staticmethod
     @njit(cache=True)
     def _first_principal_component(
-        X: NDArray[float], tol: float, max_iter: int = 100
+        X: NDArray[float],
+        tol: float,
+        v0: Optional[NDArray[float]] = None,
+        max_iter: int = 100,
     ) -> Tuple[NDArray[float], float]:
         """Compute first principal component.
+
+        Parameters
+        ----------
+        X : numpy.ndarray
+            Input data matrix with dimensions (channels x observations)
+        tol : float
+            Tolerance for convergence.
+        v0 : numpy.ndarray or None (default)
+            Initial estimate of the principal vector with dimensions
+            (channels, )
+            If None, a random initialization is used
+        max_iter : int
+            Maximum number of iterations to estiamte the first principal
+            component.
+
+        Returns
+        -------
+
+        tuple(v, eig)
+
+        v : numpy.ndarray
+            estimated principal component vector
+        eig : float
+            eigenvalue of the covariance matrix of X (channels x channels)
 
         See :footcite:t:`Roweis1997` for additional information.
         """
 
-        v = np.random.rand(X.shape[0])
-        # v = np.ones((X.shape[0],))
-        # v[::2] = -1
+        if v0 is None:  # use a random choice
+            v = np.random.rand(X.shape[0])
+        else:
+            v = v0.flatten()
+            assert v.shape[0] == X.shape[0]
         v /= np.linalg.norm(v)
 
         for _ in range(max_iter):
@@ -300,8 +328,6 @@ class AAHCluster(_BaseCluster):
             if np.linalg.norm(eig * v - s) / s_norm < tol:
                 break
             v = s / s_norm
-        # else:
-        #     logger.warn("First PC estimation: max iteration reached!")
         return v, eig
 
     # --------------------------------------------------------------------
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -43,8 +43,6 @@ class AAHCluster(_BaseCluster):
     Parameters
     ----------
     %(n_clusters)s
-    ignore_polarity : bool
-        If true, polarity is ignored when computing distances.
     normalize_input : bool
         If set, the input data is normalized along the channel dimension.
 
@@ -56,7 +54,6 @@ class AAHCluster(_BaseCluster):
     def __init__(
         self,
         n_clusters: int,
-        ignore_polarity: bool = True,
         normalize_input: bool = False,
     ):
         super().__init__()
@@ -64,9 +61,11 @@ class AAHCluster(_BaseCluster):
         self._n_clusters = _BaseCluster._check_n_clusters(n_clusters)
         self._cluster_names = [str(k) for k in range(self.n_clusters)]
 
-        self._ignore_polarity = AAHCluster._check_ignore_polarity(
-            ignore_polarity
-        )
+        # TODO : ignor_polarity=True for now.
+        # After _BaseCluster and Metric support ignore_polarity
+        # make the parameter an argument
+        # https://github.com/vferat/pycrostates/pull/93#issue-1431122168
+        self._ignore_polarity = True
         self._normalize_input = AAHCluster._check_normalize_input(
             normalize_input
         )
@@ -177,6 +176,7 @@ class AAHCluster(_BaseCluster):
         self._cluster_centers_ = maps
         self._labels_ = segmentation
         self._fitted = True
+
     # pylint: enable=arguments-differ
 
     @copy_doc(_BaseCluster.save)
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -244,6 +244,8 @@ class AAHCluster(_BaseCluster):
             to_remove = np.argmin(GEV)
             orphans = assignment == to_remove
 
+            old_cluster = cluster[:, to_remove].copy()
+
             cluster = np.delete(cluster, to_remove, axis=1)
             GEV = np.delete(GEV, to_remove, axis=0)
             assignment[assignment > to_remove] = (
@@ -260,8 +262,16 @@ class AAHCluster(_BaseCluster):
             for c in cluster_to_update:
                 members = assignment == c
                 if ignore_polarity:
+
+                    old_weight = len(new_assignment == c) / members.sum()
+
+                    sgn = np.sign(old_cluster @ cluster[:, c])
+
+                    v0 = old_weight * sgn * old_cluster + \
+                        (1. - old_weight) * cluster[:, c]
+
                     v, _ = AAHCluster._first_principal_component(
-                        data[:, members], tol, cluster[:, c]
+                        data[:, members], tol, v0
                     )
                     cluster[:, c] = v
                 else:
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -1,0 +1,372 @@
+"""Class and functions to use Atomize and Agglomerate Hierarchical Clustering
+   (AAHC)."""
+
+from functools import wraps as func_wraps
+from pathlib import Path
+from typing import Any, Optional, Tuple, Union
+
+import numpy as np
+from mne import BaseEpochs
+from mne.io import BaseRaw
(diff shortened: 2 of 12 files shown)


mscheltienne:
Thank you very much for adding this algo! I just went once through the code quickly, it looks very good. I will try to test it quickly :)

For the style, you can run `black` at the root of the repository with the command `black .` or `black pycrostates` and it will auto-format the files, fixing most of the style issues.

Also, could you add an entry in the changelog: https://github.com/vferat/pycrostates/tree/main/docs/source/dev/changes
You can add your name and a link to your website/github in `names.inc` and an entry for this PR in `latest.rst` 🙏

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+
+                    v0 = old_weight * sgn * old_cluster + \
+                        (1. - old_weight) * cluster[:, c]
+
+                    v, _ = AAHCluster._first_principal_component(
+                        data[:, members], tol, v0
+                    )
+                    cluster[:, c] = v
+                else:
+                    cluster[:, c] = np.mean(data[:, members], axis=0)
```
vferat, line 278: It seems that the mean is made on the wrong axis (`axis=0` instead of `axis=1`)

`cluster[:, c] = np.mean(data[:, members], axis=1)`

Otherwise, an error is raised we trying to fit `AAHCluster `with `ignore_polarity=False`

Would you have time to also add test(s) to cover this case ?

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+        if self.fitted:
+            n_samples = self._fitted_data.shape[-1]
+            ch_types, ch_counts = np.unique(
+                self.get_channel_types(), return_counts=True
+            )
+            ch_repr = [
+                f"{ch_count} {ch_type.upper()}"
+                for ch_type, ch_count in zip(ch_types, ch_counts)
+            ]
+            GEV = int(self._GEV_ * 100)
```
vferat, line 93: I feel like the convertion to interger could lead to missunderstanding ( GEV = 50.56% but displays 50%).
Maybe something like `GEV = f'{self._GEV_ * 100:.2f}'` would be more suitable.

I see that `ModKMeans `has the same issue. Feel free to correct it for both Classes, otherwise I'll do it in another PR.

vferat (commented):
Hey @rkobler,

Thanks a lot for this contribution, we really appreciate it.  As @mscheltienne  said, the code is very clean. This a a great addition to the library.
I checked your code with some data, and only find one/two small issues. Please have a look at my review for more details.

mscheltienne:
I have one question about the argument `ignore_polarity`. The ModKMeans always ignore polarities, right?
Why is this an option with AAHC and what is a use case of `ignore_polarity=False`?

pycrostates/cluster/__init__.py
```diff
@@ -15,5 +15,6 @@
 to segment."""
 
 from .kmeans import ModKMeans  # noqa: F401
+from .aahc import AAHCluster
```
mscheltienne, line 18:
```suggestion
from .aahc import AAHCluster  # noqa: F401
```

It's just to avoid warnings in IDEs ;)

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+    def fit(
+        self,
+        inst: Union[BaseRaw, BaseEpochs],
+        picks: Picks = "eeg",
+        tmin: Optional[Union[int, float]] = None,
+        tmax: Optional[Union[int, float]] = None,
+        reject_by_annotation: bool = True,
+        n_jobs: int = 1,
+        *,
+        verbose: Optional[str] = None,
```
mscheltienne, line 156: @vferat `verbose` should probably be defined and set in the base-class. Which will remove from both ModKMeans and this class the `fill_doc` decorator around the `fit` method and its docstring.

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+        tol: Union[int, float],
+    ) -> Tuple[NDArray[float], NDArray[int]]:
+        """Compute microstates maps."""
+        n_chan, n_frame = data.shape
+
+        cluster = data.copy()
+        cluster /= np.linalg.norm(cluster, axis=0, keepdims=True)
+
+        if normalize_input:
+            data = cluster.copy()
```
mscheltienne, line 236: Could you add a test with `normalize_input=True`?

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+        Parameters
+        ----------
+        X : numpy.ndarray
+            Input data matrix with dimensions (channels x observations)
+        tol : float
+            Tolerance for convergence.
+        v0 : numpy.ndarray or None (default)
+            Initial estimate of the principal vector with dimensions
+            (channels, )
+            If None, a random initialization is used
```
mscheltienne, line 307:
```suggestion
        v0 : numpy.ndarray or None (default)
            Initial estimate of the principal vector with dimensions
            (channels,). If None, a random initialization is used.
```

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+
+                    v, _ = AAHCluster._first_principal_component(
+                        data[:, members], tol, v0
+                    )
+                    cluster[:, c] = v
+                else:
+                    cluster[:, c] = np.mean(data[:, members], axis=0)
+                    cluster[:, c] /= np.linalg.norm(
+                        cluster[:, c], axis=0, keepdims=True
+                    )
```
mscheltienne, line 281: I was looking at how `tol` was used, and how it defines convergence. Note: I am not familiar with AAHC.

It looks like:
- `tol` is not used if `ignore_polarity=False`. That's weird, right?
- There is no information/fix possible in case the convergence was not reached. I think a warning should be emitted if convergence was not reached and `AAHCluster._first_principal_component` has a `max_iter` argument which should be forwarded to the public API, to enable the user to increase it if need be.

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+
+        Returns
+        -------
+
+        tuple(v, eig)
+
+        v : numpy.ndarray
+            estimated principal component vector
+        eig : float
+            eigenvalue of the covariance matrix of X (channels x channels)
```
mscheltienne, line 320:
```suggestion
        v : numpy.ndarray
            Estimated principal component vector.
        eig : float
            Eigenvalue of the covariance matrix of X (channels x channels)
```

I'm going to guess `pydocstyle` would complain otherwise.

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+        X : numpy.ndarray
+            Input data matrix with dimensions (channels x observations)
+        tol : float
+            Tolerance for convergence.
+        v0 : numpy.ndarray or None (default)
+            Initial estimate of the principal vector with dimensions
+            (channels, )
+            If None, a random initialization is used
+        max_iter : int
+            Maximum number of iterations to estiamte the first principal
```
mscheltienne, line 309:
```suggestion
            Maximum number of iterations to estimate the first principal
```

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,408 @@
...
+        v : numpy.ndarray
+            estimated principal component vector
+        eig : float
+            eigenvalue of the covariance matrix of X (channels x channels)
+
+        See :footcite:t:`Roweis1997` for additional information.
+        """
+
+        if v0 is None:  # use a random choice
+            v = np.random.rand(X.shape[0])
```
mscheltienne, line 326: It looks like `v0` is always provided, thus I would remove this condition and edit the arguments type-hint and docstring.

pycrostates/utils/_checks.py
```diff
@@ -67,6 +67,7 @@ def __instancecheck__(cls, other):
     "path-like": (str, Path, os.PathLike),
     "int": (_IntLike(),),
     "callable": (_Callable(),),
+    "bool": (bool,),
```
mscheltienne, line 70: Why did you need this?

`_check_type(True, (bool,))` should work directly, no need for a string wrapper.

pycrostates/cluster/aahc.py
rkobler: Thank you for the positive feedback. I have not tested the case `ignore_polarity=False`. I will fix the error and add a test case.

pycrostates/cluster/aahc.py
rkobler: Yes, I will add a test case.

pycrostates/utils/_checks.py
rkobler: You are right. I reverted this change and updated the corresponding calls in aahc.py

pycrostates/cluster/aahc.py
rkobler: Resolved in https://github.com/vferat/pycrostates/pull/92/commits/a3f7afc062b2f43cf61a9bc2b43f7af29a470524

pycrostates/cluster/aahc.py
rkobler: Resolved in https://github.com/vferat/pycrostates/pull/92/commits/1161e8eeba3a0c522d5e5577454d45423393527a

pycrostates/cluster/aahc.py
rkobler: Resolved in https://github.com/vferat/pycrostates/pull/92/commits/a3f7afc062b2f43cf61a9bc2b43f7af29a470524

pycrostates/cluster/aahc.py
rkobler: Thank you for the comment. Yes, `tol` was only used in the case `ignore_polarity=True`. In this case, AAHC requires to compute the first principal component (PC) of the cluster members. There are two options. First, power iterations (as implemented), which is fast but requires some heuristic to determine convergence. Second, conventional PCA which is slow because it requires the covariance matrix and then computes all eigenvectors/values.

In commit  https://github.com/vferat/pycrostates/pull/92/commits/a3f7afc062b2f43cf61a9bc2b43f7af29a470524, I removed the `tol` parameter from the public API. In case the fast power iteration method does not converge within a reasonable iteration (e.g., number of features = number of channels), the fallback method is used (covariance matrix + largest eigenvalue/vector).

pycrostates/cluster/aahc.py
rkobler: Resolved in commit https://github.com/vferat/pycrostates/pull/92/commits/a3f7afc062b2f43cf61a9bc2b43f7af29a470524

pycrostates/cluster/aahc.py
rkobler: Resolved in commit https://github.com/vferat/pycrostates/pull/92/commits/29fa0e53ffd2a1aefc633a32e47bec430afcb275

rkobler:
@mscheltienne 

> I have one question about the argument `ignore_polarity`. The ModKMeans always ignore polarities, right?
Why is this an option with AAHC and what is a use case of `ignore_polarity=False`?

This is a valid question. I think the best motivation for including the case `ignore_polarity=False` are scenarios where one would like to apply clustering to event-related potential (ERP) datasets. In this case, the polarity is informative.

vferat:
> @mscheltienne 
> 
> > I have one question about the argument `ignore_polarity`. The ModKMeans always ignore polarities, right?
> Why is this an option with AAHC and what is a use case of `ignore_polarity=False`?
> 
> This is a valid question. I think the best motivation for including the case `ignore_polarity=False` are scenarios where one would like to apply clustering to event-related potential (ERP) datasets. In this case, the polarity is informative.

@rkobler is right. For ERP analysis we take into account polarities whereas we ignore it for spontaneous activities ( raw / epochs).

I did not intend to include ERP microstates analysis in pycrostates at first. But it may be interesting. For now, I think we should keep the `ignore_polarity` hidden from the public API  and always use the default value `True`.

 `Metrics`, `.predict()`  are only implemented to work with ignored polarities. I may lead to wrong results if people try to use them with a cluster that do not ignore polarities.

In the meantime, I'll have a look at how much work is needed to implement changes to take into account polarities in the library.

mscheltienne:
Ok, then yes +1 to include `ignore_polarity` in ModKMeans as well in a different PR. And indeed, if for now `predict` does not work when polarities are not ignored, then let's hard-code `self._ignore_polarity = True` in `AAHC.__init__()` with a TODO item to move it back as an argument when we add support for `ignore_polartity=False` in the base-class, in ModKMeans, and in the metrics.

pycrostates/cluster/aahc.py
mscheltienne: Thank you, looks great, I love this fallback ;)

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,386 @@
...
+        normalize_input: bool,
+    ) -> Tuple[NDArray[float], NDArray[int]]:
+        """Compute microstates maps."""
+        n_chan, n_frame = data.shape
+
+        cluster = data.copy()
+        cluster /= np.linalg.norm(cluster, axis=0, keepdims=True)
+
+        if normalize_input:
+            data = cluster.copy()
```
mscheltienne, line 227:
```suggestion
        cluster = data.copy()

        if normalize_input:
            cluster /= np.linalg.norm(cluster, axis=0, keepdims=True)
```

Shouldn't it be this? Or am I missing something?

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,386 @@
...
+                        Cxx = data[:, members] @ data[:, members].T
+                        _, V = np.linalg.eigh(Cxx)
+                        v = V[:, -1]
+
+                    cluster[:, c] = v
+                else:
+                    cluster[:, c] = np.mean(data[:, members], axis=1)
+                    cluster[:, c] /= np.linalg.norm(
+                        cluster[:, c], axis=0, keepdims=True
+                    )
```
mscheltienne, line 281: One more question about this part, I'm a bit surprised by the difference between the steps when `ignore_polarity` is set to True and the steps when it is set to False. 
Could you expand a bit on those, either here, or even maybe in a couple of comments in the code?

Mostly, I'm trying to build confidence that it shouldn't be:

```
if ignore_polarity:
    # do some extra-steps

cluster[:, c] = np.mean(data[:, members], axis=1)
cluster[:, c] /= np.linalg.norm(cluster[:, c], axis=0, keepdims=True)
```

Also, do you know if there is a scikit-learn implementation of this algorithm (which would be the case `ignore_polarity=False`)? I found `sklearn.cluster.AgglomerativeClustering` but I'm not sure it's the same algo.

mscheltienne: Nevermind, @vferat convinced me.

pycrostates/cluster/aahc.py
rkobler: The version that I submitted is consistent with the reference implementation [1] (eeg_computeAAHC.m lines 15 to 19). That is, the clusters *and* data are normalized.

[1] https://www.thomaskoenig.ch/Download/EEGLAB_Microstates/Microstates1.2.zip

pycrostates/cluster/aahc.py
rkobler: I agree with @mscheltienne . Should I implement this change in this PR?

pycrostates/cluster/aahc.py
mscheltienne: I completely misread those lines, I thought this copy wasn't achieving anything, but it does replace the argument `data` with itself normalized.

pycrostates/cluster/aahc.py
mscheltienne: If you feel like it sure, else I'll do it after.

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # --------------------------------------------------------------------
+
+    @property
+    def ignore_polarity(self) -> bool:
+        """If true, polarity is ignored when computing distances.
+
+        :type: `bool`
+        """
+        return self._ignore_polarity
```
mscheltienne, line 359: This property should be removed for now, until we add proper support.

pycrostates/cluster/aahc.py
```diff
@@ -0,0 +1,397 @@
...
+
+                    v, _, converged = AAHCluster._first_principal_component(
+                        data[:, members], v0, max_iter=n_chan
+                    )
+                    if not converged:
+                        # fall back to covariance estimation
+                        # and eigenvalue computation
+                        Cxx = data[:, members] @ data[:, members].T
+                        _, V = np.linalg.eigh(Cxx)
+                        v = V[:, -1]
```
mscheltienne, line 284: Could you add a logger line at the debug level, to be able to check how often we have to fall back?
I would even add the `else` statement with another debug line to check how often we don't have to.

pycrostates/cluster/tests/test_aahc.py
```diff
@@ -0,0 +1,1358 @@
...
+    # Trust MNE .get_data() to correctly select data
+    data = epochs.get_data(picks=picks, tmin=tmin, tmax=tmax)
+    # check channels
+    assert fitted_data.shape[0] == data.shape[1]
+    # check samples
+    assert fitted_data.shape[1] == int(data.shape[0] * data.shape[2])
+
+
+def test_aahClusterMeans():
+    """Test K-Means default functionalities."""
```
mscheltienne, line 226:
```suggestion
    """Test AAHC default functionalities."""
```

pycrostates/cluster/tests/test_aahc.py
```diff
@@ -0,0 +1,1358 @@
...
+    assert aah_cluster._fitted_data is not None
+    assert aah_cluster._info is not None
+    assert aah_cluster.GEV_ is not None
+    assert aah_cluster._labels_ is not None
+
+
+def _check_unfitted(aah_cluster):
+    """
+    Checks that the aah_cluster is not fitted.
+    """
```
mscheltienne, line 188:
```suggestion
    """Check that the aah_cluster is not fitted."""
```

pycrostates/cluster/tests/test_aahc.py
```diff
@@ -0,0 +1,1358 @@
...
+    assert (dists.min(axis=0) < tol).all()
+    # ensure that all cluster centers were identified
+    assert len(set(dists.argmin(axis=0))) == sim_n_ms
+
+
+# pylint: disable=protected-access
+def _check_fitted(aah_cluster):
+    """
+    Checks that the aah_cluster is fitted.
+    """
```
mscheltienne, line 174:
```suggestion
    """Check that the aah_cluster is fitted."""
```

pycrostates/cluster/tests/test_aahc.py
```diff
@@ -0,0 +1,1358 @@
...
+    #         n_clusters=n_clusters,
+    #         ignore_polarity=None
+    #     )
+
+    # normalize_input
+    with pytest.raises(
+        TypeError, match="'normalize_input' must be an instance of bool"
+    ):
+        aahCluster_ = AAHCluster(n_clusters=n_clusters, normalize_input="asdf")
+        aahCluster_ = AAHCluster(n_clusters=n_clusters, normalize_input=None)
```
mscheltienne, line 637: Unless I am mistaken, the `pytest.raises` context manager will only catch the first matching exception and then skip the rest of the nested code. Take the `foo` function:

```
def foo(x):
    if x == 1:
        raise RuntimeError("test")
    else:
        return x
```

If you run:

```
import pytest

with pytest.raises(RuntimeError, match="test"):
    x = (foo(5))
    foo(1)
```

The variable `x` is defined at the end. But if you run:

```
import pytest

with pytest.raises(RuntimeError, match="test"):
    foo(1)
    x = (foo(5))
```

The variable `x` is not defined at the end. That means you can not test multiple calls that should raise the same error message in the same context manager, or only the first one will actually be tested.

mscheltienne (approved):
Only minor comments remain, LGTM. 
I especially like the tests on simulated data you added!

pycrostates/cluster/tests/test_aahc.py
rkobler: Thanks for pointing me on this issue. I updated the code accordingly.

mscheltienne (approved):
LGTM, thanks a lot for this amazing addition, and don't worry about the codacy workflow ;)

mscheltienne:
Good catch for `n_jobs`, @vferat I pushed an additional commit to fix the docstring, since the order of the arguments would have been wrong.

vferat:
Thx @mscheltienne for the fix !

LGTM ! @rkobler thanks a lot for the contribution!  Everything went super smooth thanks to your code quality and fast responses 🥇 

@all-contributors please add @rkobler  for code, doc, ideas and test !

PR #93, opened 2022-11-01
Add support for ERP analysis: ignore_polarity = False

### LINKED ISSUE #160: GFP equation
Looking quickly through GFP stuff, this is the equation I find in MNE:

```
np.sqrt((data**2).mean(axis=1))
```

And this is in pycrostates: https://github.com/vferat/pycrostates/blob/ff3da2b02024d5a54d09878d87d14834859d38dd/pycrostates/preprocessing/extract_gfp_peaks.py#L160

@vferat Any opinion here?

vferat:
Closes #160

As discussed in #92, we could add support for taking into account polarities in the analysis.
Here is a list of changes that need to be implemented:
- [ ] Add `self._ignore_polarity`  (`bool`) attribute to `_BaseCluster` . This attribute can only be changed using the `ignore_polarity`  argument of the `.fit()` method ( to be added).
By default during fitting,  `ignore_polarity` should be set to True for `Raw `, `Epochs`  , `ChData  ` and `False `for `Evoked `.
- [ ] `.predict` should have the same behaviour as `.fit()` : it should use  `self._ignore_polarity`.
- [ ] `ignore_polarity` value used during prediction should be store in segmentations
- [ ] `_BaseCluster.invert_polarity` should not work when `self._ignore_polarity`  is set to `False`.
- [ ] `metrics` should take into account `cluster._ignore_polarity` .
- [ ]  I/O roundtrip!

Feel free to complete this list


mscheltienne:
Adding one more bullet point: 
- [ ] I/O roundtrip!

vferat:
This pull request involves changing too many things at once in order to adapt the segmentation to ERP analysis. Furthermore, this method no longer seems to be widely used within the community.
I am closing this PR. Some features will be added via other PRs (#220 , #197)

PR #96, opened 2022-11-17
Add apply_spatial_filter

### LINKED ISSUE #6: [ENH] Add a spatial filter
Add a preprocessing function to spatialy smooth data (raw, epochs, evoked?).
Something similar to the spatial filter option in cartool
image

vferat:
Closes #6 

As discussed previously in #6, add a spatial filter similar to the one in Cartool [ref](https://www.frontiersin.org/articles/10.3389/fneur.2019.00325/full).

The current implementation:
1. Compute the adjacency matrix.
2. Compute the interpolation matrix (n_channels, n_channels).

Then for each channel:

3. Create a mask of neighbors channels ( + itself) based on the *adjacency matrix*.
4. Removes neighbors with highest and lower amplitude from mask.
5. Use mask to extract weights from the *interpolation matrix*.
6. Normalize weights.
7.  Compute current channel data by averaging neighbors data and weight.

The current implementations is working channel by channel, I add some parallelization here.

To work with `ChData`,   `mne.channels.find_ch_adjacency()` needs to inhirit from `SetChannelsMixin` and `UpdateChannelsMixin` as well as having a `.preload` and `. ch_name`s attributes.

- I currently set the `preload ` attribute to `True` during `ChData ` init.
- I added a `ch_names ` property to `ChData`

I add the two parents classes `SetChannelsMixin` and `UpdateChannelsMixin`  to `ChannelsMixin` but do no test it yet.

There is some space for improvements in channel pick and bad channel handling.


pycrostates/preprocessing/__init__.py
```diff
@@ -5,5 +5,6 @@
 
 from .extract_gfp_peaks import extract_gfp_peaks
 from .resample import resample
+from .spatial_filter import apply_spatial_filter
 
-__all__ = ("extract_gfp_peaks", "resample")
+__all__ = ("extract_gfp_peaks", "resample", "apply_spatial_filter")
```
mscheltienne, line 10:
```suggestion
__all__ = ("apply_spatial_filter", "extract_gfp_peaks", "resample")
```

Just to keep them ordered alphabetically 😅

pycrostates/io/ch_data.py
```diff
@@ -48,6 +48,8 @@ def __init__(self, data: NDArray[float], info: Union[Info, CHInfo]):
             )
         self._data = data
         self._info = info if isinstance(info, ChInfo) else ChInfo(info)
+        # Preload
+        self.preload = True
```
mscheltienne, line 52: Let's define it as a property, at least this way it can not be modified.

```
@property
def preload(self):
    """Preload required by some MNE functions."""
    return True

File: pycrostates/utils/mixin.py [hunk 3]
```diff
@@ -2,19 +2,20 @@
...
 
-class ChannelsMixin:
+    
+class ChannelsMixin(MNESetChannelsMixin, MNEUpdateChannelsMixin):
     """Channels Mixin for futur implementation."""
 
     # TODO: Maybe some part are salvageable from mne.channels.channels:
     # - UpdateChannelsMixin
     # - SetChannelsMixin
     pass
```
mscheltienne, line 17: Let's be cautious here, I don't remember which class inherits from ChannelsMixin in `pycrostates` if any; but we need to make sure this does not break any existing code (or add broken methods to them).

mscheltienne (commented):
I'll try to look at this PR too next week, just a couple of small comments for now :)

pycrostates/utils/mixin.py
vferat: yep I can confirm that.

I started to test and implement all the methods from thoses classes with `ch_data` but there is a **huge amount** of work here.
I remove the support of `ChData` instances in `apply_spatial_filter` for now.  One can apply the spatial filter before converting it to `ChData` anyway.

mscheltienne: Ok, yes let's do it this way for now.

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+from mne.parallel import parallel_func
+from mne.utils.check import _check_preload
+
+from ..utils._checks import _check_n_jobs, _check_type, _check_value
+from ..utils._docs import fill_doc
+from ..utils._logs import logger, verbose
+
+
+@fill_doc
+@verbose
```
mscheltienne, line 18: IIRC, you copied the `verbose` decorator from MNE which also fills the doc, so only one is needed.

```suggestion
@verbose
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+def apply_spatial_filter(
+    inst: Union[BaseRaw, BaseEpochs],
+    ch_type: str = "eeg",
+    exclude_bads: bool = True,
+    n_jobs: int = 1,
+    verbose=None,
+):
+    r"""Apply spatial filter.
+
+    Adapted from \ :footcite:p:`michel2019eeg`.
```
mscheltienne, line 28: `:p:` renders only as a numerical right?
IIRC, the other option, which I think would be more adapted to this sentence includes the author/year, `:n:`?

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    Apply an instantaneous filter which interpolates channels
+    with local neigbhors while removing outliers.
```
mscheltienne, line 30:
```suggestion
    with local neighbors while removing outliers.
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    The current implementation proceeds as follows:
+
+    * An interpolation matrix is computed using
+      ``mne.channels.interpolation._make_interpolation_matrix``.
+    * An ajdacency matrix is computed using
+      `mne.channels.find_ch_adjacency`.
+    * If ``exclude_bads`` is st to ``True``,
```
mscheltienne, line 37:
```suggestion
    * If ``exclude_bads`` is set to ``True``,
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+      bad channels are removed from the ajdacency matrix.
+    * For each timepoint and each channel,
+      a reduced adjacency matrix is build by removing neigbhors
+      with lowest and highest value.
+    * For each timepoint and each channel,
+      a reduced interpolation matrix is build by extracting neigbhor
```
mscheltienne, line 43:
```suggestion
      a reduced interpolation matrix is built by extracting neighbor
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    ----------
+    inst : Raw | Epochs
+        Instance to spatial filter.
+    ch_type : str
+        The channel type on which to apply spatial filter.
+        Currently only supports ``'eeg'``.
+        Default to 'eeg'.
+    exclude_bads : bool
+        If set to ``True`` bad channels will be removed
+        from the adjacency matrix and therefore no be used
```
mscheltienne, line 59:
```suggestion
        from the adjacency matrix and therefore not be used
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    inst : Raw | Epochs
+        The modified instance.
+
+    References
+    ----------
+    .. footbibliography::
+    """
+    # Checks
+    _check_type(inst, (BaseRaw, BaseEpochs))
+    _check_value(ch_type, ("eeg"), item_name="ch_type")
```
mscheltienne, line 78:
```suggestion
    _check_value(ch_type, ("eeg",), item_name="ch_type")
```

I'd be surprised it worked without the comma, IIRC, the `_check` functions don't work if you don't provide the element to check against in a tuple, even if the tuple is length 1.

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+        inst.info, ch_type
+    )
+    adjacency_matrix = adjacency_matrix.todense()
+    adjacency_matrix = np.array(adjacency_matrix)
+    if exclude_bads:
+        for c, chan in enumerate(ch_names):
+            if chan in inst.info["bads"]:
+                adjacency_matrix[c, :] = 0  # do not change bads
+                adjacency_matrix[:, c] = 0  # don't use bads to interpolate
+                print(adjacency_matrix)
```
mscheltienne, line 94: Marking this line so you don't forget a print in the release ;)

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,165 @@
...
+    # retrieve picks based on adjacency matrix
+    picks = dict(_picks_by_type(inst.info, exclude=[]))[ch_type]
+    assert ch_names == inst.ch_names
+    # retrieve channel positions
+    # TODO: warn if no sphere fit for eeg
+    pos = inst._get_channel_positions(picks)
```
mscheltienne, line 100: If you need a montage, we should check at the beginning that the instance does have one: `inst.get_montage() is not None`.

tutorials/preprocessing/plot_spatial_filter.py
```diff
@@ -0,0 +1,88 @@
...
+from pycrostates.datasets import lemon
+
+raw_fname = lemon.data_path(subject_id='010004', condition='EC')
+raw = read_raw_eeglab(raw_fname, preload=True)
+
+raw.pick('eeg')
+raw.set_eeg_reference('average')
+
+#%%
+# Spatial filter was first introduced in \ :footcite:p:`michel2019eeg`
```
mscheltienne, line 37: Same comment as before for the footcite, I think this only includes the numerical `[1]`, and the sentence would be better with a name/year + the numerical.

tutorials/preprocessing/plot_spatial_filter.py
```diff
@@ -0,0 +1,88 @@
...
+# in the context of EEG source imaging. Since then, it as also been
```
mscheltienne, line 38:
```suggestion
# in the context of EEG source imaging. Since then, it has also been 
```

tutorials/preprocessing/plot_spatial_filter.py
```diff
@@ -0,0 +1,88 @@
...
+
+raw.pick('eeg')
+raw.set_eeg_reference('average')
+
+#%%
+# Spatial filter was first introduced in \ :footcite:p:`michel2019eeg`
+# in the context of EEG source imaging. Since then, it as also been 
+# widly applied as a preprocessing step of EEG microstate analysis.
+# The goal of the preprocessing step is to reduced the effect of local transcient
+# artefacts that corrupts EEG channels.
```
mscheltienne, line 41:
```suggestion
# The goal of the preprocessing step is to reduce the effect of local transient
# artifacts that corrupt EEG channels.
```

mscheltienne:
One comment based on your first post here:

> To work with ChData, mne.channels.find_ch_adjacency() needs to inhirit from SetChannelsMixin and UpdateChannelsMixin as well as having a .preload and . ch_names attributes.

`find_ch_adjacency` operates on an Info, and it looks like it works with a ChInfo.

```
from mne.channels import find_ch_adjacency
from pycrostates.io import ChInfo

info = ChInfo(ch_names=["C1", "C2", "CP1", "CP2"], ch_types="eeg")
info.set_montage("standard_1020")
find_ch_adjacency(info, "eeg")
```

That code snippet works?

tutorials/preprocessing/plot_spatial_filter.py
```diff
@@ -0,0 +1,88 @@
+"""
+Spatial Filter
+==============
+
+This example demonstrates the effect of spatial filter on EEG data.
+"""
```
mscheltienne, line 6: To be renamed with a decimal naming schema: `XX_spatial_filter.py`.

vferat:
I rechecked the code,  `find_ch_adjacency(info, "eeg")` looks like to work.

However the struggle comes from the missing method `_get_channel_positions(picks)` that is not implemented in `ChData`.

To fix this issue, we could either:
 -  make sure `ChData` inherit from [`SetChannelsMixin`](https://github.com/mne-tools/mne-python/blob/eb14b9c55c65573a27624533e9224dcf474f6ad5/mne/channels/channels.py#L219)
 -  get channels positions using a different method, such as:

```
chs = info['chs']
ch_pos = [chs[ii]['loc'][:3] for ii in picks_idx]
```

I will make some test with `SetChannelsMixin `in the next days

mscheltienne:
Looking at the different methods in `SetChannelsMixin`, they all look like valid methods for a `ChData`. 
/!\ Valid for a `ChData` but not for a Cluster object, so I think we need to have separate inheritances between both.

vferat:
There are some errors. I was able to fix some of them. However there is still:

-  `_validate_type` that checks for `BaseRaw, BaseEpochs, or Evoked`  in `ch_data.set_eeg_reference`
-  `ch_data.anonymize` can't change the `'experimenter'` key.  Looks like it is due to the way the way the `_unlock` is handled [here](https://github.com/mne-tools/mne-python/blob/d9e40de7b47362fd0824f93c51da683806cc4a33/mne/io/meas_info.py#L2672).

I created a new branch [dev_spatial_filter_mixins](https://github.com/vferat/pycrostates/tree/dev_spatial_filter_mixins) where I'm experimenting with this issues. Feel free to edit it.

vferat:
Ready for review.
I added back support for `ChData `by creating a `CHData._get_channel_positions` method.
I shorten the tutorial which now only demonstrate the effect of spatial filter on topographies.

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+      using their reduced interpolation matrix.
+
+    Parameters
+    ----------
+    inst : Raw | Epochs | ChData
+        Instance to spatial filter.
+    ch_type : str
+        The channel type on which to apply spatial filter.
+        Currently only supports ``'eeg'``.
+        Default to 'eeg'.
```
mscheltienne, line 59:
```suggestion
```

Defaults can be found in the function signature. -1 for the duplication.

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+
+    References
+    ----------
+    .. footbibliography::
+    """
+    # Checks
+    from ..io import ChData
+
+    _check_type(inst, (BaseRaw, BaseEpochs, ChData), item_name="inst")
+    _check_value(ch_type, ("eeg",), item_name="ch_type")
```
mscheltienne, line 87:
```suggestion
    _check_type(inst, (BaseRaw, BaseEpochs, ChData), item_name="inst")
    _check_type(ch_type, (str,), item_name="ch_type")
    _check_value(ch_type, ("eeg",), item_name="ch_type")
    _check_type(exclude_bads, (bool,), item_name="exclude_bads")
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+    n_jobs = _check_n_jobs(n_jobs)
```
mscheltienne, line 88: Missing error checking for 'origin'

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+        Currently only supports ``'eeg'``.
+        Default to 'eeg'.
+    exclude_bads : bool
+        If set to ``True`` bad channels will be removed
+        from the adjacency matrix and therefore not be used
+        to interpolate neighbors. In addition, bad channels
+        will not be filtered.
+        If set to ``False``, proceed as if all channels were good.
+        Default to ``True``.
+    origin : array-like, shape (3,) | str
```
mscheltienne, line 67:
```suggestion
    origin : array of shape (3,) | str
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+    _check_preload(inst, "Apply spatial filter")
+    # check montage
+    if inst.get_montage() is None:
+        raise ValueError(
+            "No montage was set on your data, but spatial filter"
+            "can only work if digitization points for the EEG "
+            "channels are available. Consider calling inst.set_montage() "
+            "to apply a montage."
+        )
+    # remove bad channels
```
mscheltienne, line 99:
```suggestion
```

Old comment I guess?

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+    adjacency_matrix = np.array(adjacency_matrix)
+    if exclude_bads:
+        for c, chan in enumerate(ch_names):
+            if chan in inst.info["bads"]:
+                adjacency_matrix[c, :] = 0  # do not change bads
+                adjacency_matrix[:, c] = 0  # don't use bads to interpolate
+    # retrieve picks based on adjacency matrix
+    picks = dict(_picks_by_type(inst.info, exclude=[]))[ch_type]
+    info = mne.pick_info(inst.info, picks)
+    assert ch_names == info.ch_names
```
mscheltienne, line 114: This is a sanity-check and not something that is ever expected to fail due to the provided object from the user? In which case it would have to be a better error with an informative error message.

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+        Origin of the sphere in the head coordinate frame and in meters.
+        Can be ``'auto'`` (default), which means a head-digitization-based
+        origin fit.
+    %(n_jobs)s
+    %(verbose)s
+
+    Returns
+    -------
+    inst : Raw | Epochs| ChData
+        The modified instance.
```
mscheltienne, line 77:
```suggestion
    Returns
    -------
    inst : Raw | Epochs| ChData
        The instance modified in place.
        
    Notes
    -----
    This function requires a full copy of the data in memory.
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,190 @@
...
+    pos = inst._get_channel_positions(picks)
+    # test spherical fit
+    origin = _check_origin(origin, info)
+    distance = np.linalg.norm(pos - origin, axis=-1)
+    distance = np.mean(distance / np.mean(distance))
+    if np.abs(1.0 - distance) > 0.1:
+        logger.warn(
+            "Your spherical fit is poor, interpolation results are "
+            "likely to be inaccurate."
+        )
```
mscheltienne, line 125: I'm not sure we need the 'origin' argument. It's only use is to do this check, and thus for a user, the only reason to change this argument would be to get rid of this warning. 
I'm not entirely sure about what this check does, but I fee like there must be a better way to handle this.

pycrostates/preprocessing/tests/test_spatial_filter.py
```diff
@@ -0,0 +1,120 @@
...
+import numpy as np
+from mne.datasets import testing
+from mne.io.pick import _picks_by_type
+
+from pycrostates.io import ChData
+from pycrostates.preprocessing import apply_spatial_filter
+
+dir_ = testing.data_path() / "MEG" / "sample"
+fname_raw_testing = dir_ / "sample_audvis_trunc_raw.fif"
+raw_all = mne.io.read_raw_fif(fname_raw_testing, preload=True)
```
mscheltienne, line 11:
```suggestion
raw_all = mne.io.read_raw_fif(fname_raw_testing, preload=False)
```

pycrostates/preprocessing/tests/test_spatial_filter.py
```diff
@@ -0,0 +1,120 @@
...
+raw_all.info["bads"] = []
+raw_all.crop(0, 10)
```
mscheltienne, line 13:
```suggestion
raw_all.crop(0, 10)
raw.load_data()
```

pycrostates/preprocessing/spatial_filter.py
vferat: Yes this is a sanity check.

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,194 @@
...
+@fill_doc
+@verbose
+def apply_spatial_filter(
+    inst: Union[BaseRaw, BaseEpochs, CHData],
+    ch_type: str = "eeg",
+    exclude_bads: bool = True,
+    origin: Union[str, NDArray[float]] = "auto",
+    n_jobs: int = 1,
+    verbose=None,
+):
```
mscheltienne, line 29: After a second look, a couple more comments.
- `_check_origin()` is now called twice, at the beginning and after info sub-selecting, thus with 2 different info.
- the argument `ch_type` is a rename of the common `picks` argument. I think it would be better to support a regular `'picks'` argument, this way we can:
  - add future support for non-EEG channels
  - apply the spatial filter to a subset of the EEG channels
  
But I do understand that this changes somehow the structure of the function since it requires to check that the selection of channels makes sense. WDYT?

vferat: As if for now, we could remove the `ch_type` argument as the function only supports EEG channel type. 
In the case we later implement support for other channel types, the filter would still be applied to each channel type independently ( similar to `interpolate_bads`). In addition, It wouldn't really make sense to apply the filter on a subset of channels, as you will end up with an instance containing  interpolated and non interpolated channels. I think this is a really edgy case that could lead to errors if miss-used while users who would like to use this feature could use `inst.pick()` just before applying the filter. 

What do you think ?

mscheltienne:
Not forgetting about this, just lacking time ;)

tutorials/preprocessing/20_spatial_filter.py
```diff
@@ -0,0 +1,94 @@
...
+        sphere=sphere,
+        show=False,
+    )
+    mne.viz.topomap._plot_topomap(
+        raw_filter.get_data()[:, sample],
+        pos=raw_filter.info,
+        axes=axes[1, s],
+        sphere=sphere,
+        show=False,
+    )
```
mscheltienne, line 84: Why are you using the private `_plot_topomap` instead of `mne.viz.plot_topomap`?
Also, do you actually need to define the `sphere` as `np.array([0,0,0,0.1])`? `"auto"`, the default, should work?

tutorials/preprocessing/20_spatial_filter.py
```diff
@@ -0,0 +1,94 @@
...
+# in\ :footcite:t:`michel2019eeg`, with the aim to reduce the impact of
+# local transient artifacts on EEG channels. The computation of a spatial
+# filter starts with the creation of an adjacency matrix, defining whether
+# pairs of electrodes are adjacent (neighbors).
+# An adjacency matrix can be represented as a graph where each node represents
+# a sensor and each edge links adjacent sensors.
+
+adjacency, ch_names = find_ch_adjacency(info=raw.info, ch_type="eeg")
+plot_ch_adjacency(raw.info, adjacency, ch_names, kind="2d", edit=False)
+plt.show()
```
mscheltienne, line 52: I like the explanation of the internals, especially as the computation of an adjacency matrix can be erroneous. `plot_ch_adjacency` has an interactive component to edit the adjacency matrix. I think it would be good to have an additional argument to the spatial filter function:

```
from scipy.sparse import csr_matrix

def [...](ch_adjacency: Union[csr_matrix, str] = "auto"):
```

And to edit accordingly the error checking, and this tutorial to include a note about the possible editing of the adjacency matrix. For instance, in this tutorial, I would remove the link between `Cz` and `Fz`; between `Fp1` and `Fp2`, between `PO9` and `PO10`, ... yielding:

image

pycrostates/io/ch_data.py
```diff
@@ -138,6 +138,30 @@ def pick(self, picks, exclude="bads"):
...
+        """
+        picks = _picks_to_idx(self.info, picks)
+        chs = self.info["chs"]
+        pos = np.array([chs[k]["loc"][:3] for k in picks])
+        n_zero = np.sum(np.sum(np.abs(pos), axis=1) == 0)
+        if n_zero > 1:  # XXX some systems have origin (0, 0, 0)
+            raise ValueError(
+                "Could not extract channel positions for "
+                "{} channels".format(n_zero)
+            )
```
mscheltienne, line 162:
```suggestion
            raise ValueError(
                "Could not extract channel positions for "
                f"{n_zero} channels."
            )
```

pycrostates/preprocessing/spatial_filter.py
```diff
@@ -0,0 +1,189 @@
...
+
+    References
+    ----------
+    .. footbibliography::
+    """
+    _check_type(inst, (BaseRaw, BaseEpochs, CHData), item_name="inst")
+    _check_type(ch_type, (str,), item_name="ch_type")
+    _check_value(ch_type, ("eeg",), item_name="ch_type")
+    _check_type(exclude_bads, (bool,), item_name="exclude_bads")
+    origin = _check_origin(origin, inst.info)
```
mscheltienne, line 90:
```suggestion
```

`origin` is now tested twice. I think it makes more sense to test it later with the picked info.

mscheltienne:
Let's try to finalize this PR and get a new release out when we can ;)

mscheltienne:
@vferat Should the branch 'dev_spatial_filter_mixin' be deleted as well?

PR #100, opened 2022-12-01
[MNT] Run build and tests workflows on latest Python 3.11

### LINKED ISSUE #99: [MNT] Add python 3.11 to tests
Let's wait a bit more, but we should start to test for the latest version.

mscheltienne:
Closes #99

Let's see if it's green. I think the parts of MNE we use are already compatible, but probably not MNE-full because of all the viz dependencies.


mscheltienne:
`h5py` released today, all green!

vferat (approved):

PR #101, opened 2022-12-02
[MNT] Unpin matplotlib

mscheltienne:
Following #77, I've pin the version of matplotlib. 
It's not the best idea as the pin is very strict with only the latest version. It may results in conflicts with environments with many packages.

Usually, we should aim to support 4/5 minor version of our dependencies. In this PR, I reverted some of the changes from #77 to make sure older versions of matplotlib would still work.

### DIFF
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -236,7 +236,6 @@ class AAHCluster(_BaseCluster):
 
         n_steps = n_frame - n_clusters
         while cluster.shape[1] > n_clusters:
-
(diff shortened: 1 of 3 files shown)


vferat:
It will make it easier to maintain the code if we implement a  `check_cmap` function, don't you think ?

mscheltienne:
Definitely, I'll edit the PR.

mscheltienne:
@vferat Have a look :)

mscheltienne:
@vferat Good to merge?

Issue #82, opened 2022-09-27
[ENH] Auto-label/order the microstates maps


mscheltienne:
Any way to have a sort of auto-order and naming of the microstates maps by comparing them with templates?

vferat:
## Reorder

If have this old ``smart_reorder`` method that was implemented some time ago in Pycrostates:

```
   def smart_reorder(self):
        """Automaticaly reorder cluster centers.
        Returns
        ----------
        self : self
            The modfied instance.
        """
        self._check_fit()
        info = self.info
        centers = self.cluster_centers_

        template = np.array([[-0.13234463, -0.19008217, -0.01808156, -0.06665204, -0.18127315,
        -0.25741473, -0.2313206 ,  0.04239534, -0.14411298, -0.25635016,
         0.1831745 ,  0.17520883, -0.06034687, -0.21948988, -0.2057277 ,
         0.27723199,  0.04632557, -0.1383458 ,  0.36954792,  0.33889126,
         0.1425386 , -0.05140216, -0.07532628,  0.32313928,  0.21629226,
         0.11352515],
        [-0.15034466, -0.08511373, -0.19531161, -0.24267313, -0.16871454,
        -0.04761393,  0.02482456, -0.26414511, -0.15066143,  0.04628036,
        -0.1973625 , -0.24065874, -0.08569745,  0.1729162 ,  0.22345117,
        -0.17553494,  0.00688743,  0.25853483, -0.09196588, -0.09478585,
         0.09460047,  0.32742083,  0.4325027 ,  0.09535141,  0.1959104 ,
         0.31190313],
        [0.29388541,  0.2886461 ,  0.27804376,  0.22674127,  0.21938115,
         0.21720292,  0.25153101,  0.12125869,  0.10996983,  0.10638135,
         0.11575272, -0.01388831, -0.04507772, -0.03708886,  0.08203929,
        -0.14818182, -0.20299531, -0.16658826, -0.09488949, -0.23512102,
        -0.30464665, -0.25762648, -0.14058166, -0.22072284, -0.22175042,
        -0.22167467],
       [-0.21660409, -0.22350361, -0.27855619, -0.0097109 ,  0.07119601,
         0.00385336, -0.24792901,  0.08145982,  0.23290418,  0.09985582,
        -0.24242583,  0.13516244,  0.3304661 ,  0.16710186, -0.21832217,
         0.15575575,  0.33346027,  0.18885162, -0.21687347,  0.10926662,
[...]

PR #105, opened 2023-01-05
Auto-label/order microstates maps

vferat:
As discussed in #82, this PR add function(s) to automatically reorder `cluster centers`:
- `optimize_order` : find the best order to maximize auto-correlation between two cluster instances.


vferat:
@mscheltienne not sure where in the API to put the optimize_order function. Otherwise have a look !

mscheltienne (approved):
LGTM, maybe since this matches the order to another instance, we could call it `match_order` and we could add a method in the `_BaseCluster` class which takes another Cluster instance in input and matches the order directly. It would slightly shorten the user-code from:

```
from pycrostates.utils import optimize_order
from pycrostates.utils import ModKMeans

c1 = ModKMeans(...)
c2 = ModKMeans(...)
order = optimize_order(c1, c2)
c2.reorder(order)
```

to 

```

```
from pycrostates.utils import optimize_order
from pycrostates.utils import ModKMeans

c1 = ModKMeans(...)
c2 = ModKMeans(...)
c2.match_order(c1)
```

Also, if we keep the function in `utils/_clusters.py`, the public function should be added to `utils/__init__.py`?

**Comment by vferat** [comment]
I added the `'template'` argument to the `reorder_clusters`  method which uses `optimize_order`
I also moved the `utils._cluster` module to `clusters.utils` and adapted the documentation.

Let me know if you think there is still changes to do, otherwise you're free to merge.

File: pycrostates/cluster/_base.py [hunk 1]
```diff
@@ -373,19 +387,29 @@ def reorder_clusters(
...
         -----
-        The positions are 0-indexed.
         Operates in-place.
         """
         self._check_fit()
 
-        if mapping is not None and order is not None:
+        if (mapping is not None) + (order is not None) + (
+            template is not None
+        ) > 1:
```
mscheltienne, line 406:
```suggestion
        if sum(x is not None for x in (mapping, order, template) > 1:
```

A bit shorter

pycrostates/cluster/utils/utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+    _check_type(inst, (_BaseCluster,), item_name="inst")
+    inst._check_fit()
+    _check_type(template_inst, (_BaseCluster,), item_name="template_inst")
+    template_inst._check_fit()
+
+    if inst.n_clusters != template_inst.n_clusters:
+        raise ValueError(
+            "Instance and the template must have the same "
+            "number of cluster centers"
```
mscheltienne, line 50:
```suggestion
            "Instance and the template must have the same "
            "number of cluster centers."
```

mscheltienne (approved):
Very minor comments. I'll just push a couple more changes.

mscheltienne:
I added a type-hint class for Cluster, since it was not possible to type-hint to `_BaseCluster` directly (circular imports).
Alternatively, we could have used:

```
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pycrostates.clusters._base import _BaseCluster
```

which is an actual thing in the type-hint world; and an abomination in terms of syntax.

mscheltienne:
Looks like test are green, @vferat have a look and merge if happy.

PR #107, opened 2023-03-04
BUGFIX and performance improvement in AAHC clustering

rkobler:
This PR implements the following:

1. I identified and fixed a bug in the AAHC implementation in aahc.py L305 [1]. In case ignore_polarity=True, the in-product between the cluster and its members was not computed correctly. (element-wise absolute value vs. absolute value of the in-product).
2. I replaced the custom implementation of the power iteration algorithm to estimate the first principal component (if ignore_polarity=True). The new implementation uses scipy's SVD. Advantages: (i) for typical problem settings (dozens of channels) the new method is faster (ii) there are no additional dependencies and less code to maintain.

[1] https://github.com/vferat/pycrostates/blob/main/pycrostates/cluster/aahc.py#L305

### DIFF
--- a/pycrostates/cluster/aahc.py
+++ b/pycrostates/cluster/aahc.py
@@ -1,6 +1,5 @@
 """Atomize and Agglomerate Hierarchical Clustering (AAHC)."""
 
-from functools import wraps as func_wraps
 from pathlib import Path
 from typing import Any, Optional, Tuple, Union
 
@@ -15,23 +14,6 @@ from ..utils._checks import _check_type
 from ..utils._docs import copy_doc, fill_doc
 from ..utils._logs import logger
 from ._base import _BaseCluster
-
-# if we have numba, use its jit interface
-try:
-    from numba import njit
-except ImportError:
-
-    def njit(cache=False):  # pylint: disable=unused-argument
-        """Define dummy decorator for numba."""
-
-        def decorator(func):
-            @func_wraps(func)
-            def wrapper(*args, **kwargs):
-                return func(*args, **kwargs)
-
-            return wrapper
-
-        return decorator
 
 
 @fill_doc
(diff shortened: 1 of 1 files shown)


mscheltienne:
Thank you for the fix! I will take a closer look a bit later, but since CIs are happy and you wrote extensive tests for the AAHC class, I think that's all good :)

For the MNE-main workflow, #108 should fix this.
For the style workflow, could you run `black`? You just need to run from a terminal `black path/to/blackify` (with `black` installed in the python env, `pip install black`). The path can be a file or a folder to blackify the entire tree. Alternatively, most IDEs support black and you can auto-format a file from your IDE.

rkobler:
@mscheltienne thanks for your comment.
The latest commit https://github.com/vferat/pycrostates/pull/107/commits/27aa5f74689ae4eec237b9f27a4876987cc84175 resolved the style formatting error in my environment. Here is the command line output:

```bash
> black --check pycrostates/cluster/aahc.py 
All done! ✨ 🍰 ✨
1 file would be left unchanged.
```
I used black version 22.10.0.

mscheltienne:
The changes LGTM.

For `black` the workflow complains compared to your run because it is using the latest version 23.x.x. The last major version added some new formatting rules. So `pip install --upgrade black` followed by `black pycrostates/cluster/aahc.py` will fix it :)

@vferat I let you merge if happy. I would also like to cut a new release since `matplotlib` 3.7 is out for a while now and the `0.2.0` version is not compatible. `0.2.0` might also not be compatible with the latest MNE version, I remember a change on the topographic map `vlim` argument that might have affected `pycrostates` and that was likely fixed.
And it would be nice to release the additions :)

mscheltienne (approved):

rkobler:
> For `black` the workflow complains compared to your run because it is using the latest version 23.x.x. The last major version added some new formatting rules. So `pip install --upgrade black` followed by `black pycrostates/cluster/aahc.py` will fix it :)

All clear. Thanks for the quick response.

PR #109, opened 2023-03-22
Improve logger and fill_doc, copy_doc and verbose decorator

mscheltienne:
Not sure why, but it took me this week to look more into how the logger works. Learned a lot, and also found several bugs/improvements to fix.

e.g.
- Forward `__name__` and `__doc__` from the function wrapped by `@verbose`, which fixes the behavior of `fill_doc` and `copy_doc` when used with `@verbose`
- Fix how the handler log level is set
- Set the default handler log level to NOTSET, which sets it equal to the logger level
- Wrap around `sys.stdout` to fix log capture by `sphinx-gallery`
- ...

### DIFF
--- a/pycrostates/cluster/_base.py
+++ b/pycrostates/cluster/_base.py
@@ -558,6 +558,7 @@ class _BaseCluster(ABC, ChannelsMixin, ContainsMixin, MontageMixin):
         self._check_fit()
         _check_type(fname, ("path-like",), "fname")
 
+    @fill_doc
     @verbose
     def predict(
         self,
--- a/pycrostates/cluster/_base.py
+++ b/pycrostates/cluster/_base.py
@@ -24,8 +24,7 @@ from ..utils._checks import (
     _check_value,
 )
 from ..utils._docs import fill_doc
-from ..utils._logs import _set_verbose, logger
-from ..utils._logs import verbose as verbose_decorator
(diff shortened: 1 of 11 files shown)


pycrostates/cluster/_base.py
```diff
@@ -560,7 +558,7 @@ def save(self, fname: Union[str, Path]):
         self._check_fit()
         _check_type(fname, ("path-like",), "fname")
 
-    @verbose_decorator
+    @verbose
```
mscheltienne, line 561: I don't know why in this file and only this file the decorator was renamed. I'll see if a test break, but did I miss something?

pyproject.toml
```diff
@@ -141,6 +141,25 @@ match-dir = '^(?!docs|tutorials|build|dist|\.).*'
...
+    '**/pycrostates/utils/_fixes.py',
+    '**/tests/**',
+]
+
+[tool.coverage.report]
+exclude_lines = [
+    'pragma: no cover',
+    'if __name__ == .__main__.:',
+]
+precision = 2
```
mscheltienne, line 165: This setup reduces the code coverage percentage, but now it reflects more the reality. Especially removing all the `tests` files which by definition are covered at 100%.

vferat (approved):

PR #111, opened 2023-05-12
Add show_gradient parameter

### LINKED ISSUE #110: Plotting min and max value along side
In [their paper](https://www.biorxiv.org/content/10.1101/2023.04.09.536154v1)  Nikola Jajcay and Jaroslav Hlinka displayed `cluster_centers `with an additional arrow representing the min/max values (it was historically done this way).

image

I think this feature could facilitate the visual comparison of cluster centers, which is far too often subjective.
This feature should not be difficult to implement and the code for the figure above is available [here](https://github.com/jajcayn/eeg_latent_space/tree/main).

vferat:
Closes #110

### DIFF
--- a/pycrostates/viz/cluster_centers.py
+++ b/pycrostates/viz/cluster_centers.py
@@ -1,4 +1,4 @@
-"""Visualisation module for plotting cluster centers."""
+"""Visualization module for plotting cluster centers."""
 
 from typing import List, Optional, Union
 
--- a/pycrostates/viz/cluster_centers.py
+++ b/pycrostates/viz/cluster_centers.py
@@ -13,17 +13,21 @@ from numpy.typing import NDArray
 from .._typing import CHInfo
 from ..utils._checks import _check_axes, _check_type
 from ..utils._docs import fill_doc
+from ..utils._logs import logger, verbose
 
 
 @fill_doc
+@verbose
 def plot_cluster_centers(
     cluster_centers: NDArray[float],
     info: Union[Info, CHInfo],
     cluster_names: List[str] = None,
     axes: Optional[Union[Axes, NDArray[Axes]]] = None,
-    show_watershed: Optional[bool] = False,
+    show_gradient: Optional[bool] = False,
+    gradient_kwargs: Optional[dict] = None,
     *,
     block: bool = False,
+    verbose: Optional[str] = None,
     **kwargs,
 ):
     """Create topographic maps for cluster centers.
@@ -35,10 +39,15 @@ def plot_cluster_centers(
         Info instance with a montage used to plot the topographic maps.
     %(cluster_names)s
     %(axes_topo)s
-    show_watershed : bool
+    show_gradient : bool
         If True, plot a line between channel locations
         with highest and lowest values.
+    gradient_kwargs : dict
+        Additional keyword arguments to passed to
+        :func:`matplotlib.axes.Axes.plot` to plot
+        gradient line.
     %(block)s
+    %(verbose)s
     **kwargs
         Additional keyword arguments are passed to
         :func:`mne.viz.plot_topomap`.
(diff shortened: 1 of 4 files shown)


vferat:
`ModK.plot(show_gradient=True, gradient_kwargs={'color':'yellow'})`
Figure_2

@mscheltienne have a look ! Not sure about the parameter name `gradient`. Open to any suggestion

pycrostates/viz/cluster_centers.py
```diff
@@ -114,6 +142,24 @@ def plot_cluster_centers(
             ax = axes[i, j]
         # plot
         plot_topomap(center, info, axes=ax, show=False, **kwargs)
+        # Add min max vector
+        if show_gradient:
+            i_min = np.argmin(center)
+            i_max = np.argmax(center)
+            pos = mne.channels.layout._find_topomap_coords(info, picks="all")
```
mscheltienne, line 149: Maybe import only this function at the top-level? Or even import it directly in this `if` condition since it's very specific.

pycrostates/viz/tests/test_cluster_centers.py
```diff
@@ -1,14 +1,20 @@
+import logging
+
 import numpy as np
 import pytest
 from matplotlib import pyplot as plt
 from mne import create_info
 
 from pycrostates.io import ChInfo
+from pycrostates.utils._logs import logger
```
mscheltienne, line 9:
```suggestion
from pycrostates import logger
```

Defined at the top-level of the package, right?

pycrostates/viz/tests/test_cluster_centers.py
```diff
@@ -1,14 +1,20 @@
...
 from pycrostates.viz import plot_cluster_centers
 
+logger.propagate = True
+
 
-def test_plot_cluster_centers():
+def test_plot_cluster_centers(caplog):
     """Test topographic plots for cluster_centers."""
+    caplog.set_level(logging.WARNING)
```
mscheltienne, line 17: I had no idea you could do that! Good to know, usually I set the logger level before testing ^^'

pycrostates/viz/tests/test_cluster_centers.py
```diff
@@ -78,6 +84,29 @@ def test_plot_cluster_centers():
...
+
+    caplog.clear()
+    plot_cluster_centers(
+        cluster_centers,
+        info,
+        show_gradient=False,
+        gradient_kwargs={"color": "red"},
+    )
+    plt.close("all")
+    assert "`gradient_kwargs` has not effect" in caplog.text
```
mscheltienne, line 108:
```suggestion
    assert "argument 'gradient_kwargs' has not effect when" in caplog.text
```

To go with the other suggestion above.

mscheltienne (approved):
LGTM, name `gradient` is sound. Have you tried with arrows?
https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.arrow.html

pycrostates/viz/tests/test_cluster_centers.py
vferat: I just copied it from another test file in the library 👀

vferat:
> LGTM, name `gradient` is sound. Have you tried with arrows? https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.arrow.html

I just had a look at it. It's a little better, but it becomes quickly nonsense when you change the `kwargs`. Given the use that will be made of it, I think `plot `will be more than enough for the moment.

pycrostates/viz/cluster_centers.py
```diff
@@ -114,6 +142,24 @@ def plot_cluster_centers(
...
+        # Add min max vector
+        if show_gradient:
+            i_min = np.argmin(center)
+            i_max = np.argmax(center)
+            pos = _find_topomap_coords(info, picks="all")
+            if gradient_kwargs is not None:
+                ax.plot(
+                    [pos[i_min, 0], pos[i_max, 0]],
+                    [pos[i_min, 1], pos[i_max, 1]],
+                    "P-k",
```
mscheltienne, line 154: One more point, `"P-k"` is the `fmt` style of the line/marker, right? Maybe that should be passed as the default kwargs: `kwargs=dict(fmt="P-k")` or else we are risking a conflict if someone passes a different `fmt` in the kwargs.

mscheltienne (approved):
Screenshot from 2023-05-23 13-04-13

Style with the defaults. LGTM.

PR #120, opened 2023-08-23
MAINT: Future compatibility with MNE 1.6

mscheltienne:
The FIFF I/O was moved to a different module in https://github.com/mne-tools/mne-python/pull/11903

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -125,7 +125,7 @@ stages:
       displayName: 'Install package'
     - script: |
         python -m pip install --progress-bar off --force-reinstall git+https://github.com/mne-tools/mne-python
-        python -m pip install --progress-bar off --force-reinstall --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy scipy scikit-learn
+        python -m pip install --progress-bar off --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --timeout=180 numpy scipy scikit-learn matplotlib
       condition: eq(variables.PRE, 'true')
       displayName: 'Install pre-release dependencies'
     - script: mne sys_info -pd
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -142,7 +142,7 @@ stages:
     - script: python -c "import mne; mne.datasets.testing.data_path(verbose=True)"
       displayName: 'Get MNE testing datataset'
       condition: ne(variables.CACHE_MNE_RESTORED, 'true')
-    - script: pytest pycrostates --junitxml=junit/test-results.xml --cov=pycrostates --cov-report=xml
+    - script: pytest pycrostates --cov=pycrostates --cov-report=xml --cov-config=pyproject.toml
       displayName: 'Run unit tests'
     - task: PublishCodeCoverageResults@1
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -48,67 +48,67 @@ stages:
         linux 3.8:
           VM_IMAGE: 'ubuntu-latest'
           PYTHON_VERSION: '3.8'
-          MNE_DEV: false
+          PRE: false
           CACHE_MNE_PATH: $(CACHE_PATH_MNE_LINUX)
         linux 3.9:
           VM_IMAGE: 'ubuntu-latest'
           PYTHON_VERSION: '3.9'
-          MNE_DEV: false
+          PRE: false
           CACHE_MNE_PATH: $(CACHE_PATH_MNE_LINUX)
         linux 3.10:
           VM_IMAGE: 'ubuntu-latest'
           PYTHON_VERSION: '3.10'
-          MNE_DEV: false
+          PRE: false
           CACHE_MNE_PATH: $(CACHE_PATH_MNE_LINUX)
         linux 3.11:
           VM_IMAGE: 'ubuntu-latest'
           PYTHON_VERSION: '3.11'
-          MNE_DEV: false
+          PRE: false
           CACHE_MNE_PATH: $(CACHE_PATH_MNE_LINUX)
         linux 3.10 MNE-dev:
           VM_IMAGE: 'ubuntu-latest'
           PYTHON_VERSION: '3.10'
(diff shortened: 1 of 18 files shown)


mscheltienne:
Green locally on MNE 1.5 and 1.6, I think I got them all ;)

mscheltienne:
The most used private import is `_picks_to_idx` by far, and it should be made public at some point.
For x-ref: https://github.com/mne-tools/mne-python/issues/11913

vferat:
Well done! Everything seems to be working, but it's quickly going to be a headache...
I think we should set `mne>=1.6` in the dependencies of our future releases as soon as possible.

We'll keep this PR open until the MNE `1.6` release , I guess?

mscheltienne:
I think with this PR we can keep the compatibility with MNE above and below 1.6; it's not that much of an headache since the same functions are available in both versions, just in 2 different places.
I'm keeping it open for now, at least until I fix the deprecation warning of `numpy` 1.25 and add a pre-release workflow testing with `numpy` 2.0.
Then we can merge, and we **must** cut a new release before MNE 1.6 is out.

mscheltienne:
1c6799d79217b3c52e4b8e1f77ea45b126486d38 is a miss with numpy 1.25 installed. I'll try again later.

mscheltienne:
Ok, now we have the correct numpy/scipy/mne versions on that pytest run.
That came up right away:

> ##[warning]`in1d` is deprecated. Use `np.isin` instead.

And should be fixed by https://github.com/mne-tools/mne-python/pull/11994 if it's not part of our code.

azure-pipelines.yml
```diff
@@ -121,13 +121,13 @@ stages:
...
-      displayName: 'Install dependencies and MNE (stable)'
+        python -m pip install --progress-bar off --upgrade .[test]
+      displayName: 'Install package'
     - script: |
-        python -m pip uninstall -y mne
-        python -m pip install --progress-bar off --no-deps git+https://github.com/mne-tools/mne-python
-      condition: eq(variables.MNE_DEV, 'true')
-      displayName: 'Install MNE (main)'
+        python -m pip install --progress-bar off --force-reinstall git+https://github.com/mne-tools/mne-python
+        python -m pip install --progress-bar off --force-reinstall --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy scipy
```
mscheltienne, line 128:
```suggestion
        python -m pip install --progress-bar off --force-reinstall --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy scipy scikit-learn
```

azure-pipelines.yml
```diff
@@ -121,13 +121,13 @@ stages:
...
-      displayName: 'Install dependencies and MNE (stable)'
+        python -m pip install --progress-bar off --upgrade .[test]
+      displayName: 'Install package'
     - script: |
-        python -m pip uninstall -y mne
-        python -m pip install --progress-bar off --no-deps git+https://github.com/mne-tools/mne-python
-      condition: eq(variables.MNE_DEV, 'true')
-      displayName: 'Install MNE (main)'
+        python -m pip install --progress-bar off --force-reinstall git+https://github.com/mne-tools/mne-python
+        python -m pip install --progress-bar off --upgrade --pre --only-binary :all: -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --timeout=180 numpy scipy scikit-learn matplotlib
```
mscheltienne, line 128: That's the one test I changed. Instead of installing only MNE (main), I also install the pre-release version of numpy, scipy, scikit-learn and matplotlib (they need each other to work here, we can't mix a pre-release numpy with the current stable matplotlib or scikit-learn).

pycrostates/io/ch_data.py
```diff
@@ -3,10 +3,15 @@
...
-from mne.io import Info
-from mne.io.pick import _picks_to_idx, pick_info
+from mne import Info, pick_info
+from mne.utils import check_version
 from numpy.typing import NDArray
 
+if check_version("mne", "1.6"):
+    from mne._fiff.pick import _picks_to_idx
+else:
+    from mne.io.pick import _picks_to_idx
```
mscheltienne, line 13: You already saw that, but that's the syntax I used to load correctly the functions we used. I don't think it's adding a large maintenance burden, and we can keep MNE < 1.6 and >= 1.6.
When we will need to remove it, a simple grep on `from mne._fiff` or on `check_version("mne", "1.6")` will find all occurences.

pycrostates/io/fiff.py
```diff
@@ -521,13 +544,13 @@ def _read_meas_info(fid, tree):
...
             tag = read_tag(fid, pos)
-            nchan = int(tag.data)
+            nchan = int(tag.data.item())
         elif kind == FIFF.FIFF_CH_INFO:
             tag = read_tag(fid, pos)
             chs.append(tag.data)
         elif kind == FIFF.FIFF_MNE_CUSTOM_REF:
             tag = read_tag(fid, pos)
-            custom_ref_applied = int(tag.data)
+            custom_ref_applied = int(tag.data.item())
```
mscheltienne, line 553: `tag.data` is a one element numpy array.
Fix taken from MNE directly: https://github.com/mne-tools/mne-python/pull/11644

pyproject.toml
```diff
@@ -25,31 +25,31 @@ keywords = [
...
-    'numpy>=1.21',
-    'scipy',
-    'mne>=1.1.0',
+    'decorator',
+    'importlib-resources; python_version < "3.9"',
+    'jinja2',
     'joblib',
     'matplotlib',
-    'scikit-learn',
+    'mne>=1.2',
```
mscheltienne, line 48: Bumping MNE minimum requirement to 1.2, since this will be released with MNE 1.6, and keeping compatibility with the 4 last versions of our core dependencies is good practice.

mscheltienne (commented):
@vferat I added the main changes as review comments. The rest is only some clean-up, sorting, .. nothing important.
Hopefully CIs should be green now, with one bemol I did not solve. Warnings do not fail the CI, and I would have missed that numpy 1.25 deprecation warning if I did not click on the CI details.
If I'm right, warnings do fail my github action workflows with a similar configuration, so I'm not sure why it's not the case with azp here.

PR #123, opened 2023-10-18
Remove legacy setup.py

mscheltienne:
Not needed anymore with setuptools 64.0.0 supporting PEP 660 for editable installs.


mscheltienne:
@vferat FYI, that was a bad idea, `readthedocs` needs it..
Added back in #124

PR #139, opened 2023-12-22
Add epoch selection to EpochsSegmentation

### LINKED ISSUE #138: EpochsSegmentation should be indexable to retrieve the segmentation per condition
Similar to an epoch object where you can do `epochs["my condition"]`, an `EpochsSegmentation` should be indexable.
Suggested by Vincent today ;)

mscheltienne:
Closes #138

### DIFF
--- a/pycrostates/segmentation/segmentation.py
+++ b/pycrostates/segmentation/segmentation.py
@@ -1,6 +1,8 @@
 """Segmentation module for segmented data."""
 
-from typing import Optional, Union
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 from matplotlib.axes import Axes
 from mne import BaseEpochs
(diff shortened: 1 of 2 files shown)


pycrostates/segmentation/segmentation.py
```diff
@@ -173,3 +248,8 @@ def plot(
     def epochs(self) -> BaseEpochs:
         """`~mne.Epochs` instance from which the segmentation was computed."""
         return self._inst.copy()
+
+    @property
+    def metadata(self) -> Optional[DataFrame]:
+        """Epochs metadata."""
+        return self._inst.metadata
```
mscheltienne, line 255: TODO: remove and edit docstring above, we should probably not start exposing all methods from an epoch object here.

PR #148, opened 2024-01-17
Fix format of stubs file

mscheltienne:
Still a bit of experimenting and tuning with the stubs and the added UP ruff rule.


.github/workflows/stubs.yaml
```diff
@@ -31,6 +32,7 @@ jobs:
       - name: Generate stub files
         run: python tools/stubgen.py
       - name: Push stub files
+        if: github.event_name == 'push'
```
mscheltienne, line 35:
```suggestion
        if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```

PR #154, opened 2024-02-05
Add codecov token to workflow

mscheltienne:
@vferat Can you check in the settings that we do have the `CODECOV_TOKEN` correctly set?
It's now a required argunebt with `codecov/codecov-action@v4`


mscheltienne:
`Upload result --- {"result": "RequestResult(error=None, warnings=[], status_code=200, text='')"}`

LGTM

PR #158, opened 2024-02-14
Improve figure size and show argument in viz functions

### LINKED ISSUE #97: [ENH] Improve default figure size when plotting cluster centers
When you plot cluster centers, the figure has a lot of 'white' space above and below. It messes up a bit the render of the tutorials, e.g. https://pycrostates.readthedocs.io/en/latest/generated/auto_tutorials/clustering/plot_modK.html#sphx-glr-generated-auto-tutorials-clustering-plot-modk-py

The default figure is created here https://github.com/vferat/pycrostates/blob/c6af5bd6319ac601b317b48b299f46754ad09adf/pycrostates/viz/cluster_centers.py#L68-L70

It should be updated to create a figure with a more appropriate size.

### LINKED ISSUE #126: `plot_raw_segmentation` has not a `show` parameter
`plot_raw_segmentation` has not a `show` parameter 

https://github.com/vferat/pycrostates/blob/588c644eab08c83cd4a1a67b2e01f7adf475aee0/pycrostates/viz/segmentation.py#L100

mscheltienne:
Closes #97 
Closes #126 

Try to define a better figsize for cluster center plots.
Move the show argument outsides of the kwargs of the viz functions to make it explicit.
Fix ruff configuration which now requires separate `tools.ruff.lint` and `tools.ruff.format` section.
Enable ruff lint rules `A` and `B`.

### DIFF
--- a/pycrostates/_typing.py
+++ b/pycrostates/_typing.py
@@ -11,25 +11,25 @@ from numpy.random import Generator, RandomState
 from numpy.typing import NDArray
 
 
-class CHData(ABC):
+class CHData(ABC):  # noqa: B024
     """Typing for CHData."""
 
     pass
 
 
-class CHInfo(ABC):
+class CHInfo(ABC):  # noqa: B024
     """Typing for CHInfo."""
 
     pass
 
 
-class Cluster(ABC):
+class Cluster(ABC):  # noqa: B024
     """Typing for a clustering class."""
 
     pass
 
 
-class Segmentation(ABC):
+class Segmentation(ABC):  # noqa: B024
     """Typing for a clustering class."""
 
     pass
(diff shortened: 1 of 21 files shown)


mscheltienne:
Screenshot from 2024-02-14 14-45-38

Figsize looks better.
@vferat merge if happy. Also, OK to cut a new release with the recent additions?

vferat:
Renders are way better this way ! Thanks

Any reason not to include `show = plt.isinteractive() if show is None else show` in  a `_check_show`  function ?

```
def _check_show(show):
    """Check show parameter."""
    _check_type(show, (bool, None), "show")
    show = plt.isinteractive() if show is None else show
    return show
```

mscheltienne:
Done as `_ensure_valid_show`, I renamed the function because I make the difference between:
- `_check_xxx` which returns None
- `_ensure_xxx` which returns a validated object of `xxx`

vferat:
Perfect !
LGTM, and I agree that a release could be nice, a minor version `0.4.2` maybe ?

mscheltienne:
I would say 0.5.0
Nomenclature is major.minor.bugfix and I would not consider the new features as part of a bug fix release ;)

Issue #116, opened 2023-07-13
ERPs


Mattlab-hub:
Is there a way to perform the segmentation on ERPs?

mscheltienne:
@vferat Any reason why we excluded Evoked?
I guess you could bypass this by using a `ChData` object: https://pycrostates.readthedocs.io/en/latest/api/generated/pycrostates.io.ChData.html#pycrostates.io.ChData

```
evoked = Evoked(...)
ch_data = ChData(evoked.get_data(), evoked.info)
```

vferat:
Hey,

The way microstate analysis is performed on ERPs is quite different from spontaneous activity: one major difference is that polarity is not ignored for ERPs (more details here: [A Tutorial on Data-Driven Methods for Statistically Assessing ERP Topographies](https://link.springer.com/article/10.1007/s10548-013-0310-1)).

This is not yet supported by Pycrostates. There is an ongoing PR (#93) that should deal with it; however, more work is needed, especially to adapt the smoothing algorithm ([here](https://github.com/vferat/pycrostates/blob/fb3be393187faa933b3ece892ccbbea9dc4424ef/pycrostates/cluster/_base.py#L943)). Any contribution is welcomed! 😊

As for now, I can only recommend using [Cartool](https://sites.google.com/site/cartoolcommunity/) or [Ragu](https://www.thomaskoenig.ch/index.php/work/ragu) for the topographic analysis of ERPs.

Mattlab-hub:
Thanks @vferat. That is good to know. 
As for the smoothing parameters, I am actually working on a simulation that applies MS-like topographies and then injects 1/f noise. My hope is to see how different combinations of smoothing parameters affect the segmentation, compared to the known a priori segmentation used to create the simulation. The assumption is that each participant will be optimized by different smoothing parameter combinations, so this is just a tool to be used in tandem with actual participant data to help identify best smoothing parameters per subject. I do not have this code up on Git yet as I am still working out some of the bugs, but it is mostly done. I am happy to share it if you think it would be helpful.

Mattlab-hub:
Hi @vferat, just wondering if there has been any breakthroughs on getting the ERP side of pycrostates up an running? I have been having a really difficult time trying to figure out how to get my fif files converted into something Cartool or Ragu can read.

vferat:
Hey Matt,

Unfortunately, I don't know when we'll have time to work on this.

[...]

PR #161, opened 2024-02-16
[WIP] Enhance distance and gpf metrics

vferat:
#160 

- [x]  Change correlation computation to use cosine similarity
- [x]  Change gfp to either use `std `or `rms`

TODO:
- [ ]  make sure to use  correlation and gfp  whenever its needed


vferat:
For now, some test are failing due to hard coded test case working only with correlation (currently using cosine)

pycrostates/cluster/kmeans.py
```diff
@@ -317,7 +332,10 @@ def _compute_maps(
         for _ in range(max_iter):
             # Assign each sample to the best matching microstate
             activation = maps.dot(data)
```
vferat, line 334: use distance here ?

pycrostates/utils/utils.py
```diff
@@ -7,55 +7,60 @@
...
-        distances, copy=False, nan=10e300, posinf=1e300, neginf=-1e300
-    )
-    return distances
+def _distance_matrix(X, ignore_polarity=True):
+    corr = _correlation_matrix(X, ignore_polarity)
+    dist = 1 - corr
+    return dist
+
+
+def _gev(data, maps, segmentation):
```
vferat, line 60: this is only explained variance

pycrostates/cluster/kmeans.py
mscheltienne: Sounds good.

vferat:
This PR addresses also addresses #197 and #220, and brings a significant refactoring of the code. The main goal is to consolidate the distance metric into a single function, which helps avoid confusion between different distance definitions (e.g., Euclidean, correlation, etc.). This refactor also facilitates the introduction of the `ignore_polarity `parameter, which is essential for resolving issue #116.

Key improvements:

A single function now handles all distance metric calculations, making the code more maintainable and flexible.
The new structure makes it easier to implement additional distance definitions in the future, should they be needed.
The `ignore_polarity `parameter has been integrated

In my testing, the results from the tutorial builds remain consistent with the previous behavior when `ignore_polarity=True`. I will conduct a more comprehensive comparison between the current PR code and the main branch to ensure there are no discrepancies before the PR is reviewed.

Issue #176, opened 2024-04-29
Change slack in favor of MNE forum 


vferat:
As there is less and less users on _slack_, and don't have a budget for our own discourse, would it be possible to redirect _Pycrostates_ users to [MNE forum](https://mne.discourse.group/) instead of _slack_ ? Or at least remove the _slack_ channel from the documentation.

@mscheltienne ?

mscheltienne:
Sure, sounds good to me.

PR #177, opened 2024-04-29
Change slack in favor of MNE discourse

vferat:
As discussed in #176


mscheltienne:
Sneaking a fix for CIs, macOS-latest now runs on arm architecture.

Issue #195, opened 2024-09-05
GFP vs Cluster


ivano7777:
Hello,

I tried the tool following the tutorial, but applying it to MEG, instead of EEG (of course modifying a few details). I liked the individual microstates that I computed, after selecting my "meg_good_segments" this way:

epochs = mne.make_fixed_length_epochs(meg_good_segments, duration=2)  
#I tried without, results are similar. but don;t use too short epochs.
ModK = ModKMeans(n_clusters=n_microstates, random_state=42)
ModK.fit(epochs, n_jobs=4, picks = 'meg')
ModK.rename_clusters(new_names=["A", "B", "C", "D"])
ModK.plot(show_gradient=True)

Then, always according to the tutorial, for group analysis (whatever we want to use clusters or GFP) we need to compute the individual gfp. I did it, and lookes at the microstates using the gfp, and they are weird.

Questions:
1. Is it normal that they are so different? (GFP vs Cluster?)
2. What does the software do when it doesn't use the GFP for the fit? I am not sure I got it.

Thanks

vferat:
Hello @ivano7777,

Thank you for reaching out,

In group analysis scenarios, it's not always necessary to compute GFP peaks. Extracting GFP peaks can be done on when preprocessed recordings to select samples with a high signal-to-noise ratio. This approach not only yields smoother maps but also reduces the number of samples needed for K-means clustering, consequently lowering computation times.

For instance, in the tutorial [group-level-with-individual-clusters](https://pycrostates.readthedocs.io/en/latest/generated/auto_tutorials/group_level_analysis/00_group_level_with_individual_clusters.html#sphx-glr-generated-auto-tutorials-group-level-analysis-00-group-level-with-individual-clusters-py), we can bypass the GFP peaks extraction as demonstrated below:

```python
individual_cluster_centers = list()
for subject_id in subject_ids:
    # Load data
    raw_fname = lemon.data_path(subject_id=subject_id, condition=condition)
    raw = read_raw_eeglab(raw_fname, preload=True)
    raw = lemon.standardize(raw)
    raw.pick("eeg")
    raw.set_eeg_reference("average")

    # Subject level clustering
    ModK = ModKMeans(n_clusters=5, random_state=42)
    ModK.fit(raw, n_jobs=2)
    individual_cluster_centers.append(ModK.cluster_centers_)

# Assemble individual cluster centers
group_cluster_centers = np.vstack(individual_cluster_centers).T
[...]

PR #197, opened 2024-09-10
Fix gfp computation

vferat:
As discussed in https://github.com/vferat/pycrostates/issues/160 and https://github.com/vferat/pycrostates/issues/195, introduces changes to the Global Field Power (GFP) computation based on the selected channel type.

When the channel type is `eeg`, the GFP is now computed using the standard deviation (std).
When the channel type is `grad`or `mag`, the GFP is computed using the root mean square (rms).


pycrostates/preprocessing/extract_gfp_peaks.py
```diff
@@ -103,6 +117,18 @@ def extract_gfp_peaks(
     picks_all = _picks_to_idx(inst.info, inst.ch_names, none="all", exclude="bads")
     _check_picks_uniqueness(inst.info, picks)
 
+    # gfp function
+    ch_type = inst.get_channel_types(picks)[0]
```
mscheltienne, line 121: Is it safe to assume channel type uniqueness in the dataset? If so, I would write this as:

```suggestion
    ch_type = inst.get_channel_types(picks, unique=True)
    assert len(ch_type) == 1  # sanity-check
    ch_type = ch_type[0]
```

pycrostates/preprocessing/extract_gfp_peaks.py
```diff
@@ -27,10 +27,24 @@
...
+        return np.std(data, axis=0)
+
+    return compute_std
+
+
+def _rms():
+    def compute_rms(data):
+        return np.sqrt(np.mean(data**2, axis=0))
+
+    return compute_rms
```
mscheltienne, line 45: No need to nest here and you could use a dict to store the GFP functions per channel type:

```python
_GFP_FUNC: dict[str, Callable] = {
    "eeg": np.std,
    "grad": lambda x: np.sqrt(np.mean(x**2, axis=0)),
    "mag": lambda x: np.sqrt(np.mean(x**2, axis=0)),
}


_check_value(ch_type, _GFP_FUNC, "ch_type")
gfp = _GFP_FUNC[ch_type](data)
```

It removes the additional private functions and the nesting of function within, and removes the if/else logic on channel types.

pycrostates/preprocessing/extract_gfp_peaks.py
vferat: Yes it is safe due to `_check_picks_uniqueness(inst.info, picks)`

mscheltienne:
This is good to merge? @vferat

vferat:
The changes do not change numerical results (gpf peaks, gev, state gev) for EEG data using the mne sample dataset.

PR #235, opened 2025-06-05
Update spatial_filter.py

dr-r-kan:
PR: Fixed indexing issue in the _channel_spatial_filter() which led to array setting error.

Issue was only occurring on some files, which were loaded from fifs, converted from edf imports. Now runs smoothly.

### DIFF
--- a/docs/source/dev/changes/latest.rst
+++ b/docs/source/dev/changes/latest.rst
@@ -23,6 +23,7 @@ Bugs
 ~~~~
 
 - Update montage name from ``standard_1020`` to ``colin27_1005`` for mne.version >= 1.13 (:pr:`276` by `Victor Férat`_).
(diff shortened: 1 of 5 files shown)


vferat:
Hey @KishiBayes, thanks for opening this PR.

Could you provide a bit more details about the issue you encountered, and what this code change wants to achieve ?

From what I understand, there can be an issue in the current code version:

```
  max_mask = neighbors_data == np.amax(neighbors_data, keepdims=True, axis=0)
  min_mask = neighbors_data == np.amin(neighbors_data, keepdims=True, axis=0)
  keep_mask = ~(max_mask | min_mask)
```

where several neighbors can have a min/mask value and  therefore removing more than 2 neighboring values, resulting in an error. Is that it ?

I may be wrong, please don't hesitate to let me know if it is the case.

vferat:
Closes #283

PR #274, opened 2026-06-18
Fix _segment normalization to match Koenig (SmoothLabels.m) reference

Frederic-vW:
## Summary

- Replace per-time-point data normalization and per-map std normalization in `_segment` with Koenig's row L2-norm (`1/sqrt(Ne)`) on maps only, matching `NormDim(M,2)/sqrt(Ns)` in the Matlab reference (`SmoothLabels.m`).
- Fix a latent bug where in-place map normalization (`states /= std`) corrupted subsequent epoch predictions when `_segment` was called in a loop over epochs.
- Verify that `_smooth_segmentation` is structurally equivalent to the Koenig reference once maps are correctly normalised.

## Motivation

The previous std-normalization made `lambda` (the smoothing factor) incomparable to the original Pascual-Marqui/Koenig implementation. With row L2-norm normalization, `lambda` is directly transferable between implementations.

## Test plan

- [x] Confirm existing tests pass
- [ ] Verify smoothing output matches Koenig Matlab reference for same `lambda` value
- [ ] Check multi-epoch prediction (the epochs-loop bug fix)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pycrostates/cluster/_base.py
+++ b/pycrostates/cluster/_base.py
@@ -603,7 +603,7 @@ class _BaseCluster(ABC, ChannelsMixin, ContainsMixin, MontageMixin):
             provided.
         factor : int
             Factor used for label smoothing. ``0`` means no smoothing. Default to 0.
-            
+
             .. versionchanged:: 0.7
                 The effect of factor value now match :footcite:p:`Koenig2011` implementation.
         half_window_size : int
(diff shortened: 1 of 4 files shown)


vferat:
Hello @Frederic-vW,

Thanks for spotting the issue and for the fix !  

Do you know if the previous implementation give wrong microstate sequences ? Or the change is mainly to have similar lambda values across libraries ?

I had a quick look at the library and did not found any other implementation of functions using the  `std` / Z score normalization, so we should be good to go with your current implementation

pycrostates/cluster/_base.py
```diff
@@ -923,14 +927,20 @@ def _segment(
         tol: int | float,
         half_window_size: int,
     ) -> ScalarIntArray:
-        """Create segmentation. Must operate on a copy of states."""
-        data -= np.mean(data, axis=0)
```
vferat, line 927: What about data normalization ?

vferat:
@Frederic-vW 

> Fix a latent bug where in-place map normalization (states /= std) corrupted subsequent epoch predictions when _segment was called in a loop over epochs.

If I'm not wrong, applying several time
```
states -= states.mean(axis=1, keepdims=True)
states /= states.std(axis=1, keepdims=True)
```
shouldn't create any issue as data shouldn't be changed (` - 0` & `/ 1` after first run) ?

PR #276, opened 2026-06-24
[MAINT] Fix mne pre release montage name

vferat:
- Update montage name from "standard_1020" to "colin27_1005"
- Update reads the docs os version

### DIFF
--- a/pycrostates/cluster/tests/test_aahc.py
+++ b/pycrostates/cluster/tests/test_aahc.py
@@ -913,7 +913,7 @@ def test_picks_fit_predict(caplog):
 
     # create mock raw for fitting
     info_ = create_info(["Fp1", "Fp2", "CP1", "CP2"], sfreq=1024, ch_types="eeg")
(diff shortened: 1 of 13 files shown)


pycrostates/cluster/tests/test_aahc.py
```diff
@@ -913,7 +913,7 @@ def test_picks_fit_predict(caplog):
 
     # create mock raw for fitting
     info_ = create_info(["Fp1", "Fp2", "CP1", "CP2"], sfreq=1024, ch_types="eeg")
-    info_.set_montage("standard_1020")
```
mscheltienne, line 916: You might have to use `mne.utils.check_version` to use `standard_1020` for version below X and `colin27_1020` above X.

vferat:
I noticed that some docstrings containing ``Added in version`` from MNE are not change to  ``Added in version MNE``, notably in ``ContainsMixin``. Also some glitch in the return section rendering of ``inst: same as the input data``

